PR #91, opened 2024-05-29
feat: implement MUL opcode

### LINKED ISSUE #5: Opcode: MUL

Danikim01:
Closes #5 
This PR implements the code generation for the MUL opcode

### DIFF
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -94,14 +94,13 @@ fn push_push_normal_mul() {
 
 #[test]
 fn mul_wraps_result() {
-    let a = 128;
-    let b = 2;
+    let a = [0xFF; 32];
     let program = vec![
-        Operation::Push32(new_32_byte_immediate(a)),
-        Operation::Push32(new_32_byte_immediate(b)),
+        Operation::Push32(a),
+        Operation::Push32(new_32_byte_immediate(2)),
         Operation::Mul,
     ];
-    run_program_assert_result(program, 0);
+    run_program_assert_result(program, 254);
 }
 
 #[test]
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -79,3 +79,20 @@ fn push_push_add() {
 fn add_with_stack_underflow() {
     run_program_assert_revert(vec![Operation::Add]);
 }
+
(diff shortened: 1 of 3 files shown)


tests/operations.rs
```diff
@@ -79,3 +79,20 @@ fn push_push_add() {
 fn add_with_stack_underflow() {
     run_program_assert_revert(vec![Operation::Add]);
 }
+
+#[test]
+fn push_push_mul() {
+    let (a, b) = (2, 2);
```
MegaRedHand, line 85: Please make it two different numbers, so we can catch more errors.

MegaRedHand (commented):
Good work!! Please merge with main and address comments.

tests/operations.rs
```diff
@@ -79,3 +79,20 @@ fn push_push_add() {
...
+
+    let program = vec![
+        Operation::Push32(new_32_byte_immediate(a)),
+        Operation::Push32(new_32_byte_immediate(b)),
+        Operation::Mul,
+    ];
```
MegaRedHand, line 91: Please add a test for numeric overflow. When both numbers are big enough, the result should wrap around.

MegaRedHand (changes requested):
I forgot about this edge-case.

tests/operations.rs
Danikim01: what do you mean by wrap around?

tests/operations.rs
```diff
@@ -92,6 +92,18 @@ fn push_push_mul() {
     run_program_assert_result(program, a * b);
 }
 
+#[test]
+fn mul_wraps_result() {
+    let a = 128;
+    let b = 2;
+    let program = vec![
```
MegaRedHand, line 99: This example doesn't check u256 overflow, just u8 overflow. For that, we should have a number close to 2^256 - 1, and multiply it by something that overflows it. Note that you'll need to manually represent the number with an array in the operation. Here's an example: https://github.com/lambdaclass/evm_mlir/blob/3ab4e499cca565c5b10f16efc7709e0ee4b98a9f/tests/operations.rs#L106-L119

PR #94, opened 2024-05-29
feat: implement PUSH N opcodes

### LINKED ISSUE #71: Opcode: PUSH<N>

TomasArrachea:
Closes #71 

This PR implements the code generation for all the [PUSH](https://www.evm.codes/#5f) opcodes.


src/opcodes.rs
```diff
@@ -35,6 +132,196 @@ impl Operation {
             };
             let op = match Opcode::from(opcode) {
                 Opcode::ADD => Operation::Add,
+                Opcode::PUSH0 => {
+                    i += 32;
+                    Operation::Push0
+                }
```
MegaRedHand, line 138: Then here we can do `Operation::Push(BigUint::new(0));`

src/codegen/operations.rs
```diff
@@ -22,17 +22,48 @@ pub fn generate_code_for_op<'c, 'r>(
     op: Operation,
 ) -> Result<(BlockRef<'c, 'r>, BlockRef<'c, 'r>), CodegenError> {
     match op {
+        Operation::Push0 => codegen_push(context, region, [0; 32]),
```
MegaRedHand, line 25: And here we just have one arm with:

```rust
Operation::Push(value) => codegen_push(context, region, value)
```

src/opcodes.rs
```diff
@@ -20,6 +85,38 @@ impl From<u8> for Opcode {
 #[derive(Debug, Clone)]
 pub enum Operation {
     Add,
+    Push0,
+    Push1([u8; 1]),
+    Push2([u8; 2]),
+    Push3([u8; 3]),
+    Push4([u8; 4]),
```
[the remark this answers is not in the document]
TomasArrachea: Done!

src/opcodes.rs
TomasArrachea: Done!

src/codegen/operations.rs
TomasArrachea: Done!

tests/operations.rs
```diff
@@ -29,261 +30,260 @@ fn new_32_byte_immediate(value: u8) -> [u8; 32] {
     arr
 }
 
+fn new_push_op(bytes: &[u8]) -> Operation {
+    Operation::Push(BigUint::from_bytes_be(bytes.into()))
```
MegaRedHand, line 34: We need to update the other tests

tests/operations.rs
```diff
@@ -29,66 +30,297 @@ fn new_32_byte_immediate(value: u8) -> [u8; 32] {
...
+fn push32_push32_pop() {
     // Push two values to the stack and then pop once
     // The program result should be equal to the first
     // pushed value
     let (a, b) = (1, 2);
     let program = vec![
-        Operation::Push32(new_32_byte_immediate(a)),
-        Operation::Push32(new_32_byte_immediate(b)),
+        new_push_op(&new_32_byte_immediate(a)),
+        new_push_op(&new_32_byte_immediate(b)),
```
MegaRedHand, line 323: Can we simplify this a bit? We can use `BigUint::from` for unsigned integers. Also, we can just move the `BigUint::from` into the declaration of `a` and `b`:

```rust
let a = BigUint::from(1);
let program = vec![Operation::Push(a)];
```

TomasArrachea: I agree. Should we keep having tests for each pushn operation or just test for a single push in this way?

MegaRedHand: We can write a single test that runs multiple programs, one for each push opcode.

TomasArrachea: Ok. Although I think it will be better to keep the definition like this to avoid using a different crate to parse BigUInt to u8:
`
let a = 1_u8;
let program = vec![Operation::Push(BigUInt::from(a)];
`

tests/operations.rs
```diff
@@ -24,303 +24,73 @@ fn run_program_assert_revert(program: Vec<Operation>) {
...
-    let the_answer = 42;
-    let program = vec![new_push_op(&[the_answer; 20])];
-    run_program_assert_result(program, the_answer);
+    // Test for PUSH0, PUSH1, ... , PUSH32
+    for i in 0..33 {
+        let bytes = vec![the_answer; i];
+        let value = BigUint::from_bytes_be(&bytes);
+        let program = vec![Operation::Push(value)];
+        run_program_assert_result(program, the_answer);
+    }
```
MegaRedHand, line 37: We should use some non-zero number:

```rust
    let value = BigUint::from(5);
    // Test for PUSH1, ... , PUSH32
    // TODO: do PUSH0 manually?
    for i in 0..32 {
        let shifted_value = value << (i * 8);
        let program = vec![Operation::Push(value)];
        run_program_assert_result(program, shifted_value.try_into().unwrap());
    }
```

tests/operations.rs
MegaRedHand: You can use [`TryFrom`](https://docs.rs/num-bigint/latest/num_bigint/struct.BigUint.html#impl-TryFrom%3CBigUint%3E-for-u8) for that.

Example:

```rust
let x = BigUint::from(5);
let b: u8 = TryFrom::<BigUint>::try_from(x).unwrap();
// This is the same as:
let c: u8 = x.try_into().unwrap();
```

tests/operations.rs
TomasArrachea: Ok, but using shift it's a problem because it makes the values too big to parse to u8 and check it, no?

MegaRedHand: Yes. In that case, you can manually truncate it to that size:

```rust
let expected_result: u8 = (shifted_value % 256_u32).try_into().unwrap()
```

tests/operations.rs
```diff
@@ -23,56 +24,61 @@ fn run_program_assert_revert(program: Vec<Operation>) {
...
 #[test]
-fn push32_once() {
-    let the_answer = 42;
-    let program = vec![Operation::Push32([the_answer; 32])];
-    run_program_assert_result(program, the_answer);
+fn push_once() {
+    let value = BigUint::from(5_u8);
+
+    // For PUSH0
+    let shifted_value = BigUint::ZERO << (32 * 8);
```
MegaRedHand, line 32: Nit. No need for the shift

src/opcodes.rs
```diff
@@ -193,11 +227,201 @@ impl Operation {
             let op = match Opcode::from(opcode) {
                 Opcode::ADD => Operation::Add,
                 Opcode::MUL => Operation::Mul,
+                Opcode::PUSH0 => {
+                    i += 32;
```
MegaRedHand, line 231: Are you sure this is OK? I thought `PUSH0` increased the `pc` by 1 only (done outside the match).

src/opcodes.rs
```diff
@@ -193,11 +227,201 @@ impl Operation {
...
+                    Operation::Push(BigUint::ZERO)
+                }
+                Opcode::PUSH1 => {
+                    i += 1;
+                    let x = bytecode[i..(i + 1)].try_into().unwrap();
+                    i += 31;
```
MegaRedHand, line 237: Same here.

MegaRedHand (changes requested):

src/opcodes.rs
TomasArrachea: You are right!

PR #95, opened 2024-05-29
feat: implement SAR opcode

### LINKED ISSUE #28: Opcode: SAR

ricomateo:
Closes #28 
This PR implements the code generation for the [SAR](https://www.evm.codes/#1d?fork=cancun) opcode.


src/codegen/operations.rs
```diff
@@ -177,6 +178,50 @@ fn codegen_pop<'c, 'r>(
...
+    stack_push(context, &ok_block, result)?;
+
+    Ok((start_block, ok_block))
+}
+
+// fn integer_constant(context: &MeliorContext, value: [u8; 32]) -> Attribute {
+//     let str_value = BigUint::from_bytes_be(&value).to_string();
+//     // TODO: should we handle this error?
+//     Attribute::parse(context, &format!("{str_value} : i256")).unwrap()
+// }
```
MegaRedHand, line 224: We should remove this.

src/codegen/operations.rs
```diff
@@ -177,6 +178,50 @@ fn codegen_pop<'c, 'r>(
...
+) -> Result<(BlockRef<'c, 'r>, BlockRef<'c, 'r>), CodegenError> {
+    let start_block = region.append_block(Block::new(&[]));
+    let context = &op_ctx.mlir_context;
+    let location = Location::unknown(context);
+
+    // Check there's enough elements in stack
+    let flag = check_stack_has_at_least(context, &start_block, 2)?;
+
+    // Create REVERT block
+    let revert_block = region.append_block(generate_revert_block(context)?);
```
MegaRedHand, line 193: We can remove this, in favor of `op_ctx.revert_block`

src/codegen/operations.rs
```diff
@@ -27,6 +27,7 @@ pub fn generate_code_for_op<'c>(
         Operation::Pop => codegen_pop(op_ctx, region),
         Operation::Jumpdest { pc } => codegen_jumpdest(op_ctx, region, pc),
         Operation::Push(x) => codegen_push(op_ctx, region, x),
+        Operation::Sar => codegen_sar(op_ctx, region),
```
MegaRedHand, line 30: Please mark the opcode as implemented in the readme

src/opcodes.rs
```diff
@@ -0,0 +1,56 @@
+#[derive(Debug)]
```
MegaRedHand, line 1: We can remove this file

tests/operations.rs
```diff
@@ -140,6 +140,56 @@ fn pop_with_stack_underflow() {
...
+#[test]
+fn push_push_sar() {
+    let (value, shift) = (BigUint::from(2_u8), BigUint::from(1_u8));
+    let program = vec![
+        Operation::Push(value),
+        Operation::Push(shift.clone()),
+        Operation::Sar,
+    ];
+    let expected_result = 2 >> 1;
+    //let expected_result: u8 = value.into() >> shift.into();
```
MegaRedHand, line 152:
```suggestion
    let expected_result = (value >> shift).try_into().unwrap();
```

tests/operations.rs
```diff
@@ -140,6 +140,56 @@ fn pop_with_stack_underflow() {
...
+}
+
+#[test]
+fn sar_with_stack_underflow() {
+    let program = vec![Operation::Sar];
+    run_program_assert_revert(program);
+}
+
+#[test]
+fn sar_with_negative_value_preserves_sign() {
```
MegaRedHand, line 163: We should include a test for the case where the shift is greater than 256.

MegaRedHand (changes requested):

tests/operations.rs
ricomateo: Done!

juanbono (changes requested):
From the docs https://mlir.llvm.org/docs/Dialects/ArithOps/#arithshrsi-arithshrsiop:

```
The shrsi operation shifts an integer value of the first operand to the right by the value of the second operand. The first operand is interpreted as signed, and the second operand is interpreter as unsigned. The high order bits in the output are filled with copies of the most-significant bit of the shifted value (which means that the sign of the value is preserved). If the value of the second operand is greater or equal than bitwidth of the first operand, then the operation returns poison.
```

You should add a check for that last reason to avoid poisoned values.

ricomateo:
> From the docs https://mlir.llvm.org/docs/Dialects/ArithOps/#arithshrsi-arithshrsiop:
> 
> ```
> The shrsi operation shifts an integer value of the first operand to the right by the value of the second operand. The first operand is interpreted as signed, and the second operand is interpreter as unsigned. The high order bits in the output are filled with copies of the most-significant bit of the shifted value (which means that the sign of the value is preserved). If the value of the second operand is greater or equal than bitwidth of the first operand, then the operation returns poison.
> ```
> 
> You should add a check for that last reason to avoid poisoned values.

In those cases the operation should be reverted?

Oppen:
> @juanbono in those cases the operation should be reverted?

No. According to the EVM docs, there is no such requirement, so you'll have to either check before and keep the minimum between the actual value and the width or check for poison later and replace with `0x00..00` or `0xff..ff` depending on the sign bit.

Oppen:
You can use [minui](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithminui-arithminuiop) on the `shift` value and a constant `255` to achieve that, unconditionally, before applying the shift. You would use the result instead of `shift` for that. That way you can satisfy the mlir requirements while implementing the EVM semantics. This should be doable without adding extra blocks.

README.md
```diff
@@ -16,6 +16,7 @@ An EVM-bytecode to machine-bytecode compiler using MLIR and LLVM.
 
 1. (0x01) ADD
 1. (0x02) MUL
+1. (0x1D) SAR
```
azteca1998, line 19: I've seen in other PRs they're asking to keep this list sorted by the opcode ID.

azteca1998 (approved):

ricomateo:
> You can use [minui](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithminui-arithminuiop) on the `shift` value and a constant `255` to achieve that, unconditionally, before applying the shift. You would use the result instead of `shift` for that. That way you can satisfy the mlir requirements while implementing the EVM semantics. This should be doable without adding extra blocks.

Done!

README.md
ricomateo: Nice catch!

tests/operations.rs
Oppen: Let's also add one for a positive number (shift resulting in 0).

ricomateo: Done!

Oppen (approved):

MegaRedHand (approved):
LGTM

src/program.rs
```diff
@@ -210,22 +210,23 @@ impl From<u8> for Opcode {
...
+            x if x == Opcode::SWAP7 as u8 => Opcode::SWAP7,
+            x if x == Opcode::SWAP8 as u8 => Opcode::SWAP8,
+            x if x == Opcode::SWAP9 as u8 => Opcode::SWAP9,
+            x if x == Opcode::SWAP10 as u8 => Opcode::SWAP10,
+            x if x == Opcode::SWAP11 as u8 => Opcode::SWAP11,
+            x if x == Opcode::SWAP12 as u8 => Opcode::SWAP12,
+            x if x == Opcode::SWAP13 as u8 => Opcode::SWAP13,
+            x if x == Opcode::SWAP14 as u8 => Opcode::SWAP14,
+            x if x == Opcode::SWAP15 as u8 => Opcode::SWAP15,
+            x if x == Opcode::SWAP16 as u8 => Opcode::SWAP16,
```
Oppen, line 229: Wow nice spot!

juanbono (approved):

PR #96, opened 2024-05-30
DUP<N> opecode

### LINKED ISSUE #72: Opcode: DUP<N>

FrancoGiachetta:
Closes #72 
This PR handles the DUP<N> instruction


src/opcodes.rs
```diff
@@ -18,7 +52,9 @@ impl From<u8> for Opcode {
 #[derive(Debug, Clone)]
 pub enum Operation {
     Add,
+    Sub,
     Push32([u8; 32]),
+    DupN(u32),
```
MegaRedHand, line 57: We should just call it `Dup` since, apparently, [there's a future DUPN opcode](https://ethereum-magicians.org/t/introduce-opcodes-b0-dupn-and-b1-swapn-ignore/12765)). Also, let's add a comment that only numbers from 1 to 16 are supported.

src/utils.rs
```diff
@@ -149,6 +149,64 @@ pub fn stack_push<'ctx>(
     Ok(())
 }
 
+// Returns a copy of the nth value of the stack
+pub fn get_nth_from_stack<'ctx>(
+    context: &'ctx MeliorContext,
+    block: &'ctx Block,
+    nth: u32,
+) -> Result<Value<'ctx, 'ctx>, CodegenError> {
+    let uint256 = IntegerType::new(context, 256);
```
MegaRedHand, line 158: We should add a `debug_assert!(nth < MAX_STACK_SIZE as u32);` for good measure.

tests/operations.rs
```diff
@@ -75,6 +76,77 @@ fn push_push_add() {
     run_program_assert_result(program, a + b);
 }
 
+#[test]
+fn dup1_once() {
+    let program = vec![
+        Operation::Push32(new_32_byte_immediate(10)),
+        Operation::Push32(new_32_byte_immediate(32)),
+        Operation::DupN(1),
+    ];
```
MegaRedHand, line 85: We should add a `Pop` just for good measure.
```suggestion
        Operation::DupN(1),
        Operation::Pop,
    ];
```

src/codegen/operations.rs
```diff
@@ -69,6 +72,39 @@ fn codegen_push<'c, 'r>(
...
+fn codegen_dup<'c, 'r>(
+    codegen_ctx: CodegenCtx<'c>,
+    region: &'r Region<'c>,
+    nth: u32,
+) -> Result<(BlockRef<'c, 'r>, BlockRef<'c, 'r>), CodegenError> {
+    let start_block = region.append_block(Block::new(&[]));
+    let context = &codegen_ctx.mlir_context;
+    let location = Location::unknown(context);
+
+    //TODO check nth is not 0
```
MegaRedHand, line 84: For these kinds of sanity checks you can use `debug_assert!`.

src/utils.rs
```diff
@@ -149,6 +149,64 @@ pub fn stack_push<'ctx>(
...
+            LoadStoreOptions::default(),
+        ))
+        .result(0)?;
+
+    // Decrement stack pointer
+    let old_stack_ptr = block
+        .append_operation(llvm::get_element_ptr(
+            context,
+            stack_ptr.into(),
+            DenseI32ArrayAttribute::new(context, &[-(nth as i32 - 1)]),
```
MegaRedHand, line 188: Since nth starts from 1, the `-1` at the end shouldn't be needed.

src/codegen/operations.rs
```diff
@@ -66,6 +69,41 @@ fn codegen_push<'c, 'r>(
     Ok((start_block, ok_block))
 }
 
+fn codegen_dup<'c, 'r>(
+    codegen_ctx: CodegenCtx<'c>,
+    region: &'r Region<'c>,
+    nth: u32,
```
Oppen, line 75: You can use `NonZeroU32` and be safe.

FrancoGiachetta: Thanks! I'll try that then.

Oppen: Also, from checking the context I see there is a restricted range for which this function should be called, and that it is internal only. I suggest simply adding the following at the top of the function body instead:
```rust
debug_assert!(nth > 0 && nth <= 16);
```

src/codegen/operations.rs
```diff
@@ -69,6 +71,41 @@ fn codegen_push<'c, 'r>(
...
+    let start_block = region.append_block(Block::new(&[]));
+    let context = &op_ctx.mlir_context;
+    let location = Location::unknown(context);
+
+    //TODO check nth is not 0≈
+    // Check there's enough elements in stack
+    let flag = check_stack_has_at_least(context, &start_block, nth)?;
+
+    // Create REVERT block
+    let revert_block = region.append_block(generate_revert_block(context)?);
```
Oppen, line 88: You can use the revert block from context directly now.

src/utils.rs
```diff
@@ -149,6 +149,65 @@ pub fn stack_push<'ctx>(
...
+    let value = block
+        .append_operation(llvm::load(
+            context,
+            old_stack_ptr.into(),
+            uint256.into(),
+            location,
+            LoadStoreOptions::default(),
+        ))
+        .result(0)?
+        .into();
```
Oppen, line 206: I think this can be simplified by simply reading relative to the stack pointer.

tests/operations.rs
```diff
@@ -73,6 +73,63 @@ fn push_stack_overflow() {
...
+        Operation::Dup(11),
+        Operation::Dup(10),
+        Operation::Dup(13),
+        Operation::Dup(15),
+        Operation::Dup(14),
+        Operation::Dup(16),
+    ];
+
+    run_program_assert_result(program, 6);
+}
```
Oppen, line 125: This doesn't really test the intermediate steps. Maybe you can do something like:
```rust
#[rstest]
#[case(1)]
#[case(2)]
// ...
#[case(16)]
fn dup_nth(#[case] nth: usize) {
    let mut program = Vec::from_iter((0..16u8).rev().map(|x| Operation::Push(BigUint::from(x)));
    program.push(Operation::Dup(nth));
    run_program_assert_result(nth - 1);
}
```
There may be some miscalculations but that's the gist.

src/utils.rs
FrancoGiachetta: I wouldn't know how to do that. However, I've modified that method at the SWAP<N> branch to also return the address from the stack. Would you still say it is okay to make this change?

FrancoGiachetta: I've just changed the method I was telling you about, also to avoid possible conflicts with main now since I've merged.

Oppen (approved):

PR #123, opened 2024-05-31
feat: implement JUMPI opcode

### LINKED ISSUE #63: Opcode: JUMPI
See #101

The implementation should call `OperationCtx::add_jump_op` to add a jump into the jumptable block, if the condition holds true.

ricomateo:
Closes #63
This PR implements the code generation for the [JUMPI](https://www.evm.codes/#57?fork=cancun) opcode.


src/codegen/operations.rs
```diff
@@ -420,3 +421,65 @@ fn codegen_jumpdest<'c>(
...
+        .append_operation(arith::constant(
+            context,
+            integer_constant_from_i64(context, 0i64).into(),
+            location,
+        ))
+        .result(0)?
+        .into();
+
+    // compare  condition > 0  to convert condition from u256 to 1-bit signless integer
+    // TODO: change this maybe using arith::trunci
```
Oppen, line 463: The spec says any non-zero value results in taking the branch, so truncating would be incorrect if, say, `b == 2`, since truncation of `0b00...10` to one bit results in `0b0`.
For now I think it's better to call `check_is_zero` and use the negation of that to pick branches, or something akin to that, if we want to simplify. Otherwise, leave as is.

tests/operations.rs
```diff
@@ -269,3 +269,66 @@ fn jumpdest() {
...
+    let pc = BigUint::from(7_u8);
+    let condition = BigUint::from(1_u8);
+    let program = vec![
+        Operation::Push(condition),
+        Operation::Push(pc),
+        Operation::Jumpi,
+        Operation::Jumpdest { pc: 83 },
+    ];
+    run_program_assert_revert(program);
+}
```
Oppen, line 334: We need a similar test that ensure non-taken branches to bad jump destinations don't result in a revert, as the spec says it is only an error when the branch is taken. The code as implemented seems to behave correctly at a glance, but we have no test to make sure this doesn't change.

src/codegen/operations.rs
ricomateo: I figured out I could do this by changing the comparison from `condition > 0` to `condition != 0`.
In that way the value is converted and also any non-zero value is taken as `true`.

tests/operations.rs
ricomateo: Done!

Oppen (approved):

Issue #24, opened 2024-05-24
Opcode: NOT


igaray:

mahmudsudo:
taking on this issue

mahmudsudo:
hi  i need workflow approval for my pull request

PR #126, opened 2024-05-31
feat : Not opcode and updated readme

mahmudsudo:
1. #24 Not Opcode implemented with tests
2. readme updated to remove error when running command


src/codegen/operations.rs
```diff
@@ -596,3 +598,47 @@ fn codegen_jumpdest<'c>(
 
     Ok((landing_block, landing_block))
 }
+
+// from the understanding of the not operator , A xor 1 == Not A
```
MegaRedHand, line 602: This is correct only for boolean values. Here we're working with 256-bit integers. For them, bitwise NOT is equivalent to `A xor 0xff...ff` (i.e. xor-ing against 256 set bits)

MegaRedHand (changes requested):
Hey there, and thanks again for your contribution! I left you some comments. Also, CI seems not to be running because of conflicts with main.

src/codegen/operations.rs
```diff
@@ -596,3 +598,58 @@ fn codegen_jumpdest<'c>(
...
+        flag,
+        &ok_block,
+        &op_ctx.revert_block,
+        &[],
+        &[],
+        location,
+    ));
+
+    let mut mask: [u8; 32] = [0; 32];
+   mask[31] = 0xff;
```
MegaRedHand, line 2049: The mask should be `[0xff; 32]` (the 256 bits set) for it to work.

src/codegen/operations.rs
```diff
@@ -596,3 +598,58 @@ fn codegen_jumpdest<'c>(
...
+    let lhs = stack_pop(context, &ok_block)?;
+    // let one = ok_block
+    //     .append_operation(arith::constant(
+    //         context,
+    //         integer_constant_from_i64(context, 1).into(),
+    //         location,
+    //     ))
+    //     .result(0)?
+    //     .into();
```
MegaRedHand, line 2058: We can remove this

MegaRedHand (commented):
Left you some comments. Once they're addressed I think it'll be ready to merge 😄
Please move the opcode from "not implemented" to "implemented" on the readme.

src/codegen/operations.rs
mahmudsudo: done

mahmudsudo: setting the mask to [0xff;32] makes the test fail whilst 
`
 let mut mask: [u8; 32] = [0; 32];
   mask[31] = 0xff;
`
passes all tests

mahmudsudo:
#216 reimplements this in line with the changes to the codebase

PR #127, opened 2024-05-31
SHR opcode

### LINKED ISSUE #27: Opcode: SHR

FrancoGiachetta:
Close #27 
This PR handles the implementation of the SHR opcode


Oppen:
The same `minui` trick used in `SAL` should work to simplify `SHR`.

FrancoGiachetta:
Checking!

FrancoGiachetta:
I've run into a problem. The thing is that a flag in order to produce the conditional branch depending on the value of shift. Since the append_operation(...) returns an OperationRef<_ , _>, and I need a Value<_ , _>, how should I proceed?

Oppen:
> I've run into a problem. The thing is that a flag in order to produce the conditional branch depending on the value of shift. Since the append_operation(...) returns an OperationRef<_ , _>, and I need a Value<_ , _>, how should I proceed?

Ahh, I see, we need the conditional because the poison comes into play when `shift >= 256` meaning the `minui` trick works only for `255` and that preserves the high bit instead of making everything `0`. One way to work around that is to just shift twice, but that would be hacky and probably requires branching anyway.
Let's keep it as it is. We can always come back to it in a different PR if we find a better way.

Oppen:
But just to be clear, the proposal was to avoid the conditional block altogether.
A way would be to zero-extend and use `minui` with 256 and then truncate back tho. But it piles up arcane stuff and it may make it hard to understand later.

src/program.rs
```diff
@@ -28,6 +28,8 @@ pub enum Opcode {
     // NOT = 0x19,
     BYTE = 0x1A,
     // SHL = 0x1B,
+    SHR = 0x1C,
+    // SAR = 0x1D,
     // SHR = 0x1C,
```
Oppen, line 33: We should remove this comment. Also, the `SAR` change shouldn't be there.

Oppen (approved):

PR #128, opened 2024-06-01
feat: SLT opcode implemented

mahmudsudo:
#17 
SLT opcode implemented with appropriate tests

### DIFF
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -424,6 +424,20 @@ fn test_slt_signed_less_than() {
 
     run_program_assert_result(program, 0);
 }
+#[test]
+fn test_slt_both_signed() {
+    let mut a = BigUint::from(3_u8);
+    a.set_bit(255, true);
+    let mut  b = BigUint::from(2_u8);
+    a.set_bit(255, true);
+    let program = vec![
+        Operation::Push(a.clone()),
+        Operation::Push(b.clone()),
+        Operation::Slt,
+    ];
+
+    run_program_assert_result(program, 0);
+}
 
 #[test]
 fn test_slt_signed_greater_than() {
@@ -449,7 +463,7 @@ fn test_slt_equal() {
         Operation::Push(b.clone()),
         Operation::Slt,
     ];
-    run_program_assert_result(program, 0);
+    run_program_assert_result(program, 1);
 }
 
 #[test]
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -426,17 +426,16 @@ fn test_slt_signed_less_than() {
 }
 #[test]
 fn test_slt_both_signed() {
-    let mut a = BigUint::from(3_u8);
-    a.set_bit(255, true);
-    let mut  b = BigUint::from(2_u8);
-    b.set_bit(255, true);
+    let  a = BigUint::from((!3_u8 + 1 )as u8);//-3
+    let   b = BigUint::from((!2_u8 + 1 ) as u8 );//-2
+ 
     let program = vec![
         Operation::Push(a.clone()),
         Operation::Push(b.clone()),
         Operation::Slt,
     ];
 
-    run_program_assert_result(program, 1);
+    run_program_assert_result(program, 0);
 }
 
 #[test]
@@ -463,7 +462,7 @@ fn test_slt_equal() {
         Operation::Push(b.clone()),
         Operation::Slt,
     ];
-    run_program_assert_result(program, 1);
+    run_program_assert_result(program, 0);
 }
 
 #[test]
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -426,8 +426,8 @@ fn test_slt_signed_less_than() {
 }
 #[test]
 fn test_slt_both_signed() {
-    let  a = BigUint::from((!3_u8 + 1 )as u8);//-3
-    let   b = BigUint::from((!2_u8 + 1 ) as u8 );//-2
+    let  a = BigUint::from(!3_u8 + 1 );//-3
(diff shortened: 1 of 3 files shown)


Oppen (changes requested):
Thanks for the PR! Can you add tests for cases where both operands are negative?

mahmudsudo:
there seems to e an issue when the two operands are negative , i have tried converting from BigInt to BigUint to no avail

Oppen:
Can you describe the issue in more detail? Note the `set_bit(255, true)` is how you (currently) build a negative number. Also, since it's 2's complement, what you're really expressing is `u256::MIN + N` where `N` is the `BigUint` you are applying the operation to.

mahmudsudo:
#[test]
fn test_slt_both_signed() {
    let mut a = BigUint::from(3_u128);
    a.set_bit(255, true);
    let mut  b = BigUint::from(2_u128);
    b.set_bit(255, true);
    let program = vec![
        Operation::Push(a.clone()),
        Operation::Push(b.clone()),
        Operation::Slt,
    ];

    run_program_assert_result(program, 0);
}

this fails even though -3 is less than -2

Oppen:
Said a different way: if what you want is `-3` and `-2`, what you need to do is flip all the bits and add 1, as that's negation in 2's complement.

tests/operations.rs
```diff
@@ -383,3 +383,90 @@ fn addmod_with_zero_denominator() {
...
+        Operation::Push(b.clone()),
+        Operation::Slt,
+    ];
+
+    run_program_assert_result(program, 0);
+}
+#[test]
+fn test_slt_both_signed() {
+    let  a = BigUint::from((!3_u8 + 1 )as u8);//-3
+    let   b = BigUint::from((!2_u8 + 1 ) as u8 );//-2
```
Oppen, line 430: This also doesn't work as expected. `!3_u8` is `0xfc`, which is then interpreted as `0x00..fc`, a positive number.
Try building it from the `[0xffu8; 32]` array instead, it will be easier. `[0xff, ..., 0xfd]` should be `-3` and `[0xff, ..., 0xfc]` should be `-2`. You can call `from_bytes_be` on it.

mahmudsudo: on following this paradigm the test fails

mahmudsudo: #[test]
fn test_slt_both_signed() {
    // let  a = BigUint::from(!3_u8 + 1 );//-3
    // let   b = BigUint::from(!2_u8 + 1 ) ;//-2
    let mask = [0xffu8; 32];
    let mut a = mask.clone();
    let mut b = mask.clone();
    a[31]= 0xfd;
    b[31] = 0xfc;

    let program = vec![
        Operation::Push(BigUint::from_bytes_be(&a)),
        Operation::Push(BigUint::from_bytes_be(&b)),
        Operation::Slt,
    ];

    run_program_assert_result(program, 0);
}

FrancoGiachetta: Hi! May I suggest to use the Sub opcode on the test? Doing 0 - <number> would return you the negative. Doing it this way should work.

Oppen: > Hi! May I suggest to use the Sub instruction on the test? You can add a step before the checking that would perform 0 - num, with each number. Since the Sub instruction has been tested, doing it this way should work.

This is a valid option, but I think it's better to isolate what we're testing here. Otherwise later changes may break the test and we would need more debugging to figure out whether we broke `SLT` or `SUB`.

mahmudsudo: > Now, let's look at the operation order. We have `b = - 2` and `a = -3`. We push `a`, we push `b`, we execute `SLT`, which pops `b` and then pops `a` (because we remove from the top of the stack), then performs `a < b` and `a = -3 < -2 = b` is true, so we push `1`. The test as written should fail because it's expecting `-3 < -2` to push `0` instead.

but this same order works for other tests

mahmudsudo:
#188  already implemented it

tests/operations.rs
Oppen: Let's go case by case.
`test_slt_unsigned_greater_than`: tests `(b = 1 < 2 = a) = 1`, which is correct, as `1 < 2` is true.
`test_slt_unsigned_less_than`: tests `(b = 2 < 0 = a) = 0`, which is correct, as `2 < 0` is false.
`test_slt_signed_less_than`: tests `(b = 2 < i256::MIN + 3 = a) = 0`, which is correct, as `2 < i256::MIN + 3` is false (`i256::MIN + 3` is negative).
`test_slt_both_signed` (current version): tests `(b = 253 < 254 = a) = 0` which is correct (though it tests the wrong thing), as `253 < 254` is false.
`test_slt_both_signed` (version in the comment): tests `(b = -3 < -2 = a) = 0` which is incorrect, because `-3 < -2` is true, so the right side should be `1`.
`test_slt_signed_greater_than`: tests `(b = i256::MIN + 3 < 2 = a) = 1`, which is correct, because `i256::MIN + 3` is negative and thus `b < 2`.
`test_slt_equal`: tests for equal values and the program returns `0` because none is less than the other, so it is correct (and doesn't depend on ordering).
`test_slt_stack_underflow`:  tests overflow.

Note in the old and new versions of `test_slt_both_signed`, we are expecting a `0` in the written tests, but one has positive numbers with the lesser one at the right and other has negative numbers with the lesser one (larger modulus) at the left, so `SLT` evaluates to true, and that's what we should expect.

MegaRedHand:
Someone on our team thought the issue was free and took it. Sorry! We'll make sure it doesn't happen again. I can assign #185 to you if you want.

PR #142, opened 2024-06-03
feat: load executables dynamically, add memory, and implement RETURN opcode

### LINKED ISSUE #78: Opcode: RETURN

### LINKED ISSUE #108: Definition and implementation of Runtime/executor
This issue is a constraint of the issue #107. We need to solve this first before implementing E2E tests.

MegaRedHand:
Closes #78
Closes #108

This PR migrates `main.rs` code to dynamically load and execute the compiled code, passing a context that allows it to call syscalls and interact with the compiler side. In future PRs we'll move this code into some wrapper object and present a nicer API for this calls. The new syscall handler logic will allow us to implement storage and call functionalities, and the memory logic added will allow us to implement the MSTORE and MLOAD opcodes.


azteca1998 (approved):

src/utils.rs
```diff
@@ -547,6 +550,57 @@ pub fn check_if_zero<'ctx>(
     Ok(flag.into())
 }
 
+pub(crate) fn extend_memory<'c>(
```
juanbono, line 553: I think it will be useful to add some docs here so it can serve as an example of a syscall implementation.

src/syscall_handler.rs
```diff
@@ -0,0 +1,63 @@
+use std::ffi::c_void;
```
juanbono, line 1: I think a module doc can be useful here to explain the different moving parts

src/syscall_handler.rs
```diff
@@ -0,0 +1,63 @@
...
+use melior::ExecutionEngine;
+
+#[derive(Debug, Default)]
+pub struct SyscallContext {
+    pub memory: Vec<u8>,
+    result: Option<(usize, usize)>,
+}
+
+impl SyscallContext {
+    pub fn result(&self) -> &[u8] {
```
juanbono, line 12: same here, some docs can be useful to newcomers.

src/syscall_handler.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Syscall implementations
+impl SyscallContext {
+    fn write_result(&mut self, offset: usize, bytes_len: usize) {
+        self.result = Some((offset, bytes_len));
+    }
+
+    fn extend_memory(&mut self, new_size: usize) {
+        println!("Extending memory to {new_size}");
```
juanbono, line 25: here we can use tracing instead of println

src/context.rs
```diff
@@ -388,7 +423,63 @@ fn generate_stack_setup_block<'c>(
...
+
+    // Declare the stack pointer and base pointer globals
+    let body = module.body();
+    let res = body.append_operation(llvm_mlir::global(
+        context,
+        MEMORY_PTR_GLOBAL,
+        ptr_type,
+        location,
+    ));
+    assert!(res.verify());
```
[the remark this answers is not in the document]
juanbono: he is initializing it to 0

juanbono (changes requested):
see comments

src/context.rs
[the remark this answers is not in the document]
MegaRedHand: It's intended. I'm not initializing it because, for any memory access, we'll first check there's enough memory and set this global to the result of `extend_memory` anyways.

src/syscall_handler.rs
MegaRedHand: Good catch! I actually forgot to remove this print 😅

src/codegen/operations.rs
```diff
@@ -1365,6 +1348,58 @@ fn codegen_pc<'c>(
...
+        &[],
+        &[],
+        location,
+    ));
+
+    let offset_u256 = stack_pop(context, &ok_block)?;
+    let size_u256 = stack_pop(context, &ok_block)?;
+
+    let offset = ok_block
+        .append_operation(arith::trunci(offset_u256, uint32.into(), location))
```
Oppen, line 1380: Is the truncation specified? Why did we choose uint32 rather than platform native?

src/lib.rs
```diff
@@ -290,3 +279,80 @@ pub fn compile_binary(
...
+
+            args.extend(&["-o", &output_filename]);
+
+            args.extend(&["-L/lib/../lib64", "-L/usr/lib/../lib64", "-lc", "-O1"]);
+
+            args.extend(objects.iter().map(|x| x.as_str()));
+
+            args
+        }
+        #[cfg(target_os = "windows")]
```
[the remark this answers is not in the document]
Oppen: In fact, the `cfg!` macro as used [here](https://github.com/lambdaclass/evm_mlir/pull/142/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R341-R349) should suffice, considering no OS-specific functions are called.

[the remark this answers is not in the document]
Oppen: (The advantage would be just having an `else` case)

src/syscall.rs
```diff
@@ -0,0 +1,128 @@
...
+/// function to be callable from the generated code.
+impl SyscallContext {
+    pub extern "C" fn write_result(&mut self, offset: u32, bytes_len: u32) {
+        self.result = Some((offset as usize, bytes_len as usize));
+    }
+
+    pub extern "C" fn extend_memory(&mut self, new_size: u32) -> *mut u8 {
+        let new_size = new_size as usize;
+        if new_size > self.memory.len() {
+            // TODO: check for OOM
```
Oppen, line 57: You can do this with [try_reserve](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.try_reserve).

src/codegen/operations.rs
MegaRedHand: I truncated the number here for simplicity. I'm not sure how we would represent a `usize` in MLIR, and we'd probably run into some weird issues. The max size for the EVM memory is ~4MB according to the block gas limit, so we'll probably never reach it.

MegaRedHand: I added a comment

src/syscall.rs
MegaRedHand: Done

src/syscall.rs
```diff
@@ -0,0 +1,181 @@
...
+//!
+//! ### Adding a new syscall
+//!
+//! New syscalls should be implemented by adding a new method to the [`SyscallContext`]
+//! struct (see [`SyscallContext::write_result`] for an example). After that, the syscall
+//! should be registered in the [`register_syscalls`] function, which will make it available
+//! to the generated code. Afterwards, the syscall should be declared in
+//! [`mlir::declare_syscalls`], which will make the syscall available inside the MLIR code.
+//! Finally, the function can be called from the MLIR code like a normal function (see
+//! [`mlir::write_result_syscall`] for an example).
```
Oppen, line 17: Maybe we can make a helper function for this? It seems like many steps for the implementer to keep in mind. Maybe have `register_syscalls` do the `mlir::declare_syscalls`?

MegaRedHand: The thing is both are different steps of the whole workflow, and I think it's better to have them split like this to avoid mixing it in more complex ways. Besides, I think having all of this in a single module should make it simple enough, right?

juanbono (approved):

PR #144, opened 2024-06-03
feat: opcode POP (gas cost)

### LINKED ISSUE #134: Opcode: POP (gas cost)

ricomateo:
Closes #134 
This PR adds gas consumption to the [POP](https://www.evm.codes/#50?fork=cancun) opcode.

### DIFF
--- a/src/codegen/operations.rs
+++ b/src/codegen/operations.rs
@@ -968,7 +968,7 @@ fn codegen_pop<'c, 'r>(
     // Check there's at least 1 element in stack
     let flag = check_stack_has_at_least(context, &start_block, 1)?;
 
-    let gas_flag = consume_gas(context, &start_block, 3)?;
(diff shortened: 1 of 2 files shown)


src/codegen/operations.rs
```diff
@@ -968,11 +968,18 @@ fn codegen_pop<'c, 'r>(
     // Check there's at least 1 element in stack
     let flag = check_stack_has_at_least(context, &start_block, 1)?;
 
+    let gas_flag = consume_gas(context, &start_block, 3)?;
```
Oppen, line 971: According to the docs it consumes `2`.

Oppen (changes requested):

src/codegen/operations.rs
ricomateo: Nice catch!

Oppen (approved):

PR #194, opened 2024-06-05
feat: implement dynamic gas for memory extensions

### LINKED ISSUE #193: Compute dynamic gas cost of MSTORE & MSTORE8

TomasArrachea:
Closes #193 

This PR implements the code generation for the dynamic gas calculations used when the memory is extended. Also, now the memory is only extended when necessary (when `size + offset > memory size`).


src/codegen/operations.rs
```diff
@@ -1474,9 +1474,79 @@ fn codegen_mload<'c, 'r>(
         .result(0)?
         .into();
 
-    let memory_ptr = extend_memory(op_ctx, &ok_block, required_size)?;
+    // Get current memory size
+    let memory_size_ptr = ok_block
```
MegaRedHand, line 1478: Let's move this check inside `extend_memory`, along with the gas consumption. Same for the other ops. Also, let's move all gas calculation to the end, so we only update the counter and check for OOG once.

src/utils.rs
```diff
@@ -606,47 +666,169 @@ pub fn check_if_zero<'ctx>(
...
+    // memory_size_word = (memory_byte_size + 31) / 32
+    // memory_cost = (memory_size_word ** 2) / 512 + (3 * memory_size_word)
+    //
+    //
     let context = op_ctx.mlir_context;
     let location = Location::unknown(context);
+    let uint32 = IntegerType::new(context, 32).into();
 
-    let ptr_type = pointer(context, 0);
+    let thirty_one = block
```
MegaRedHand, line 683: Let's just use the number literal in the name. Same for the rest.
```suggestion
    let constant_31 = block
```

src/utils.rs
```diff
@@ -606,47 +666,169 @@ pub fn check_if_zero<'ctx>(
...
+            uint32.into(),
+            location,
+            LoadStoreOptions::default(),
+        ))
+        .result(0)?
+        .into();
 
-    let res = block.append_operation(llvm::store(
+    // Compare current memory size and required size
+    let extension_flag = check_is_greater_than(context, block, memory_size, required_size)?;
```
MegaRedHand, line 797: This should check `required_size > memory_size`, right?

src/utils.rs
```diff
@@ -606,47 +666,169 @@ pub fn check_if_zero<'ctx>(
...
+        extension_flag,
+        &extension_block,
+        &no_extension_block,
+        &[],
+        &[],
+        location,
+    ));
+
+    // Extend memory
+    let memory_ptr = op_ctx.extend_memory_syscall(&extension_block, required_size, location)?;
```
MegaRedHand, line 812: Can we move this call to after consuming gas?

src/utils.rs
```diff
@@ -693,6 +920,7 @@ pub(crate) fn return_result_from_stack(
     block: &Block,
     reason_code: ExitStatusCode,
     location: Location,
+    region: &Region<'_>,
```
MegaRedHand, line 923: nit. Can we reorder this to `op_ctx, region, block, ...` like in the other utils?

tests/operations.rs
```diff
@@ -105,7 +105,10 @@ fn test_return_with_gas() {
         Operation::Push((1, 2_u8.into())),
         Operation::Return,
     ];
-    run_program_assert_result(program, &[0]);
+    let dynamic_gas = 3;
```
MegaRedHand, line 108: Can we compute this here? Maybe adding a helper function.

tests/operations.rs
```diff
@@ -2167,3 +2177,49 @@ fn mload_not_allocated_address() {
...
+        Operation::Push0, // offset
+        Operation::Mload,
+    ];
+    let dynamic_gas = 3_i64;
+    let needed_gas = gas_cost::PUSH0 + gas_cost::MLOAD + dynamic_gas;
+    run_program_assert_gas_exact(program, needed_gas as _);
+}
+
+#[test]
+fn mload_gas_cost_with_memory_extension2() {
```
MegaRedHand, line 2217: Can we add one where the offset is so big we OOG?

tests/operations.rs
```diff
@@ -2167,3 +2177,49 @@ fn mload_not_allocated_address() {
...
+        Operation::Push((1_u8, BigUint::from(31_u8))), // offset
+        Operation::Mstore8,
+    ];
+    let dynamic_gas = 3_i64;
+    let needed_gas = gas_cost::PUSHN * 2 + gas_cost::MSTORE8 + dynamic_gas;
+    run_program_assert_gas_exact(program, needed_gas as _);
+}
+
+#[test]
+fn mload_gas_cost_with_memory_extension() {
```
MegaRedHand, line 2206: Please add tests for `RETURN` and `REVERT`

MegaRedHand (changes requested):

MegaRedHand (approved):
There's a warning from clippy, but otherwise LGTM.

PR #207, opened 2024-06-06
feat: change default return and revert to use `write_result` syscall

MegaRedHand:
This PR changes the default return (the last block) and `revert_block` to use the `write_result` syscall to mimic the expected behavior. As part of this, tests where changed to check the whole top of the stack. With this, some bugs were encountered, and failing tests were ignored. They will be fixed in other PRs.

### DIFF
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -657,7 +657,6 @@ fn push_push_normal_mul() {
 }
 
 #[test]
-// #[ignore]
 fn mul_wraps_result() {
     let a = BigUint::from_bytes_be(&[0xFF; 32]);
     let program = vec![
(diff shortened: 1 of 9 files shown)


tests/operations.rs
```diff
@@ -2068,72 +2142,5 @@ fn mload_not_allocated_address() {
         Operation::Push((1_u8, BigUint::from(32_u8))), // offset
         Operation::Mload,
     ];
-    run_program_assert_result(program, 0_u8);
-}
-
-#[test]
-fn check_initial_memory_size() {
```
MegaRedHand, line 2075: These were just moved.

juanbono (approved):
Really nice!

PR #209, opened 2024-06-07
Global variables for call data size and call data pointer declared 

Danikim01:
This PR declares the global variables for call data size and call data pointer


MegaRedHand (approved):
LGTM

PR #210, opened 2024-06-07
Add error handling to bytecode parsing

### LINKED ISSUE #201: Make `Program::from_bytecode` return errors on invalid opcodes
Currently, it just panics. It would be useful to make it return a `Result` in those cases.

JulianVentura:
Closes #201 
This PR adds error handling to bytecode parsing, so we can have information about what opcodes we are missing to implement in order to run a certain program.


MegaRedHand (approved):
Great work!

PR #211, opened 2024-06-07
test: add fibonacci example as e2e test

MegaRedHand:


src/program.rs
```diff
@@ -328,6 +328,62 @@ pub enum Operation {
     Mstore8,
 }
 
+impl Operation {
+    pub fn to_bytecode(&self) -> Vec<u8> {
```
juanbono, line 332: can we just return a [u8]?

juanbono (changes requested):
I think it can be better to not return a Vec<u8>

src/program.rs
MegaRedHand: No, because the length is arbitrary.

juanbono (approved):

PR #213, opened 2024-06-07
feat: SLOAD Opcode

### LINKED ISSUE #60: Opcode: SLOAD
To start, we can model storage as a `HashMap` or `BTreeMap` inside the `SyscallContext`.

FrancoGiachetta:
Closes #60 
This PR handles the implementation of the SLOAD opcode


src/syscall.rs
```diff
@@ -244,6 +245,20 @@ impl<'c> SyscallContext<'c> {
         }
     }
 
+    pub extern "C" fn read_storage(&mut self, stg_key: &U256, stg_value: &mut U256) {
+        let address = self.env.tx.caller;
+        let mut combined = [0u8; 32];
+
+        combined[..16].copy_from_slice(&stg_key.hi.to_be_bytes());
+        combined[16..].copy_from_slice(&stg_key.lo.to_be_bytes());
+        let key = EU256::from_big_endian(&combined);
```
TomasArrachea, line 254: I think it may be better to do it like this:
` let key = (EU256::from(stg_key.hi)) << 128) + stg_key.lo;`

FrancoGiachetta: I've tried that way but it doesn't work for some reason.

src/syscall.rs
```diff
@@ -17,6 +17,7 @@
 //! [`mlir::write_result_syscall`] for an example).
 use std::ffi::c_void;
 
+use ethereum_types::U256 as EU256;
```
MegaRedHand, line 20: This should be imported from the `primitives` module

tests/evm.rs
```diff
@@ -650,3 +650,70 @@ fn chainid_stack_overflow() {
     let env = Env::default();
     run_program_assert_halt(program, env);
 }
+
+#[test]
+fn sload_gas_consumtion() {
```
MegaRedHand, line 655: typo
```suggestion
fn sload_gas_consumption() {
```

MegaRedHand (approved):
Great work!

PR #232, opened 2024-06-11
Fix some bugs and update operations tests

### LINKED ISSUE #214: Fix ignored tests
We currently have some failing tests in `tests/operations.rs` (the ones marked `#[ignore]`). We want to fix them.

Related issues:
- [ ] #206

JulianVentura:
Closes #214 
This PR updates and fixes some ignored tests under `tests/operations.rs`. It also fixes some bugs on opcodes implementations under `codegen/operations.rs`


src/utils.rs
```diff
@@ -299,11 +299,27 @@ pub fn stack_push<'ctx>(
...
 
     let uint256 = IntegerType::new(context, 256);
+    let uint1 = IntegerType::new(context, 1);
+
+    // NOTE:
+    // Here we are extending the Value's size to a 256 bits word size.
+    // There are some operations, for instance, the comparisson ops like GT, EQ, SGT, which return
+    // a Value result of size 1 bit.
+    // We must extend those operations to 32 bytes = 256 bits padding with zeros in order to have a clean return value
+    // Otherwise, the return value may be filled with trash.
```
MegaRedHand, line 309: We should move the `extui` to each of those ops instead of adding it here.

tests/operations.rs
```diff
@@ -162,15 +162,30 @@ fn push_twice() {
...
+        That means we cannot use the hole 1024 word size. Our usable stack limit drops to 1023 elements.
+        If we didn't want to return the top of the stack value, then we would be able to use 1024 push instructions.
+    */
     let stack_top = BigUint::from(88_u8);
 
-    // Push 1024 times
-    let program = vec![Operation::Push((1_u8, stack_top.clone())); 1024];
+    // Push 1023 times
+    let program = vec![Operation::Push((1_u8, stack_top.clone())); 1023];
     run_program_assert_stack_top(program, stack_top);
```
MegaRedHand, line 175: Let's change this to use `run_program_assert_result` instead, which doesn't add the `RETURN` operation, to make it explicit. Also, we don't need to check the result (it should be `&[]`).

src/program.rs
```diff
@@ -165,10 +166,27 @@ pub enum Opcode {
...
 pub struct ParseError(Vec<OpcodeParseError>);
 
+impl fmt::Debug for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let opcodes: Vec<_> = self.0.iter().map(|x| format!("{:02X}", x.0)).collect();
+        writeln!(f, "The following opcodes could not be parsed: ")?;
+        writeln!(f, "{:#?}", opcodes)?;
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
JulianVentura: Hmm actually I changed this because with the `derive(Debug)` the log is not very clear. Actually, it's pretty confusing. Do you still want it like it was?

tests/operations.rs
```diff
@@ -573,18 +579,18 @@ fn sdiv_signed_division_1() {
     let mut b = BigUint::from(0_u8);
     b.set_bit(253, true);
 
-    //r = a / b = [1, 1, 1, 1, ....., 1, 1, 0, 0]
+    //r = a / b = [1, 1, 1, 1, ....., 1, 1, 0, 0] = -4
     //If we take the lowest byte
     //r = [1, 1, 1, 1, 1, 1, 0, 0] = 252 in decimal
-    let expected_result: u8 = 252_u8;
+    let expected_result = biguint_256_from_bigint(BigInt::from(-4_i8));
```
MegaRedHand, line 585: Can we compute this as `a / b` with `BigInt`s before converting?

src/utils.rs
JulianVentura: Hmm that was my first solution, but I noticed that would mean some code repetition. Also, this problem may appear in other operations (we may miss one too), so I think this is better because its an open solution. It should fix all those operations, even not yet implemented ones, so I think it can potentially save us some future bugs. Also, for clarification. I think that the MLIR output would be the same, since the `if` condition is being processed at MLIR code generation time. What do you think?

tests/operations.rs
```diff
@@ -1692,28 +1709,31 @@ fn stop() {
...
+    run_program_assert_stack_top(program, expected_result);
 }
 
-// TODO: fix this test
 #[test]
-#[ignore]
 fn exp_with_overflow_should_wrap() {
-    let a = 3_u8;
-    let b = 256_u32;
+    let a = BigUint::from(2_u8);
```
MegaRedHand, line 1726: Can you change the base to some non-multiple of 2, so the result is non-zero?

src/program.rs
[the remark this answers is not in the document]
MegaRedHand: Yes. The `Debug` implementation doesn't need to be pretty.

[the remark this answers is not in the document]
JulianVentura: Ok

src/utils.rs
MegaRedHand: I think it'll bring more headaches than solutions. We already implemented all comparisons, so no more `uint1` pushes should be added, and putting the complexity of extending the input inside the `stack_push` function will make future refactors (like #224) more difficult.

Aside from that, we can add an `assert!` that checks the value being pushed is a `uint256` to avoid future errors.

JulianVentura: Ok, I'll make the change. Do you have something in mind for that assert you mention? I think it will be pretty similar to the `if` clause I mentioned. Something like `assert!(value.r#type().eq(&uint256.into()))`

MegaRedHand: No, just something similar to the if condition. Also, let's use `debug_assert!` instead of `assert!`.

MegaRedHand (approved):
Great catch! 🎣

PR #255, opened 2024-06-13
feat: allow configuring the opt level on executor

MegaRedHand:


src/main.rs
```diff
@@ -24,7 +24,7 @@ fn main() {
         .compile(&program.unwrap(), &output_file)
         .expect("failed to compile program");
 
-    let executor = Executor::new(&module);
+    let executor = Executor::new(&module, Default::default());
```
juanbono, line 27: I think it is a good idea to allow the user to set the  the opt-level here

MegaRedHand: Done!

juanbono (approved):

PR #258, opened 2024-06-13
feat: implement OPCODE Origin

### LINKED ISSUE #32: Opcode: ORIGIN

Danikim01:
closes #32 
This PR implements the opcode Origin


src/syscall.rs
```diff
@@ -31,6 +31,35 @@ pub struct U256 {
...
+}
+
+impl U256 {
+    pub fn from_be_bytes(bytes: [u8; 32]) -> Self {
+        let hi = u128::from_be_bytes(bytes[0..16].try_into().unwrap());
+        let lo = u128::from_be_bytes(bytes[16..32].try_into().unwrap());
+        U256 { hi, lo }
+    }
+
+    pub fn store_h160_in_u256(&mut self, value: &ethereum_types::H160) {
```
JulianVentura, line 48: I think the name `copy_from_address` would be more clear for this function. So, I would also change the `value` type to `Address`. And remember you should import the primitive types from `src/primitives.rs`

src/syscall.rs
```diff
@@ -31,6 +31,35 @@ pub struct U256 {
     pub hi: u128,
 }
 
+impl std::fmt::Display for U256 {
```
JulianVentura, line 34: Do we need to implement `Display`?

src/syscall.rs
```diff
@@ -191,6 +220,11 @@ impl<'c> SyscallContext<'c> {
         self.env.tx.data.len() as u32
     }
 
+    pub extern "C" fn get_origin(&self, address: &mut U256) {
+        let aux: &ethereum_types::H160 = &self.env.tx.caller;
```
JulianVentura, line 224: I'm not sure, but I think that the `tx.caller` corresponds to the `CALLER` opcode. Here you are implementing `ORIGIN`.

tests/evm.rs
```diff
@@ -114,6 +114,36 @@ fn fibonacci_example() {
     assert_eq!(number, 55_u32.into());
 }
 
+#[test]
+fn test_opcode_origin() {
```
JulianVentura, line 118: Can we add a test (or change this one) to match the example given in [the EVM Codes playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code=%27ORIGIN%27_)? I have done something similar for `CALLER` opcode. Maybe that's useful to you.

azteca1998 (approved):

azteca1998 (approved):

PR #259, opened 2024-06-13
feat: add Database trait

### LINKED ISSUE #244: Add `Database` trait and use that instead of `Db`
The trait should be the same as revm's:

https://github.com/bluealloy/revm/blob/8742fc7e779e70214a66a262b37a2b90025b3429/crates/primitives/src/db.rs#L11-L26

TomasArrachea:
Closes #244 .

This PR adds a `Database` trait, and implements it for the `Db` struct. Also, it adds an `EvmBuilder`.


MegaRedHand (approved):

src/db.rs
```diff
@@ -1,19 +1,176 @@
...
+            }
+        }
+        Ok(U256::default())
+    }
+
+    fn block_hash(&mut self, number: U256) -> Result<B256, Self::Error> {
+        match self.block_hashes.entry(number) {
+            std::collections::hash_map::Entry::Occupied(entry) => Ok(*entry.get()),
+            std::collections::hash_map::Entry::Vacant(entry) => Ok(B256::default()),
+        }
```
MegaRedHand, line 92: This should use [the Entry API](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert) instead of a `match`.

MegaRedHand (changes requested):

src/db.rs
```diff
@@ -1,19 +1,187 @@
 #![allow(unused)]
 use ethereum_types::{Address, U256};
```
JulianVentura, line 2: You can import those types from `src/primitives.rs`

src/db.rs
```diff
@@ -1,19 +1,187 @@
...
+    pub bytecode_hash: B256,
 }
 
+type B256 = U256;
+
 #[derive(Clone, Debug, Default)]
 pub struct Db {
-    accounts: HashMap<Address, AccountInfo>,
+    accounts: HashMap<Address, DbAccount>,
+    contracts: HashMap<B256, Bytecode>,
```
JulianVentura, line 21: Maybe we could use `Bytes` from `src/primitves.rs` instead of `Bytecode` here. Otherwise, we could define the `Bytecode` type inside `src/primitives.rs`, maybe like this: `pub type Bytecode = Bytes`

MegaRedHand (approved):
LGTM

PR #267, opened 2024-06-14
refactor evm

### LINKED ISSUE #231: Update `Evm::transact` with `Db`
Depends on #230
`Evm::transact` should fetch tx data from the configured `Env`, and fetch the bytecode to execute from the `Db`. As part of this change, we need to remove the `Program` from the `Evm`, and update all tests.

TomasArrachea:
Closes #231 

This PR changes the way the Evm acceses the bytecode. Instead of the Program being stored inside the Evm struct, now the bytecode is stored inside the db.


tests/ef_tests.rs
```diff
@@ -138,9 +142,10 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
         let Some(account) = unit.pre.get(&to) else {
             return Err("Callee doesn't exist".into());
         };
-        let env = Env::default();
-        let program = Program::from_bytecode(&account.code)?;
-        let mut evm = Evm::new(env, program);
+        let mut env = Env::default();
+        env.tx.transact_to = TransactTo::Call(to);
+        let db = Db::with_bytecode(to, Bytecode(Into::<Vec<u8>>::into(account.code.clone())));
```
JulianVentura, line 147: I think that this line is a bit complex. Maybe if the `Bytecode` type were of type `Bytes` it would be easier. If you want to preserve the name `Bytecode`, you can create a struct that wraps a `Bytes` and implement any method you want for it.

JulianVentura: nit: If I'm creating a new struct instance from some other type, I like to call that method `from_something`. In this case, `from_bytecode`.

src/db.rs
```diff
@@ -23,8 +30,18 @@ pub struct Db {
 }
 
 impl Db {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn with_bytecode(address: Address, bytecode: Bytecode) -> Self {
+        let mut db = Db::default();
+        let mut hasher = Keccak256::new();
+        hasher.update(bytecode.as_slice());
+        let hash = B256::from_big_endian(&hasher.finalize());
```
MegaRedHand, line 37: Do we need to specify it's big endian?

tests/ef_tests.rs
MegaRedHand: I agree we should move to using bytes for the `Bytecode` object (maybe just alias `Bytecode = Bytes`). And I'll add that `with_bytecode` would be more useful if it received `self` and modified that. That way you can chain multiple calls to it.

src/lib.rs
```diff
@@ -51,8 +48,19 @@ impl Evm<Db> {
...
+
+        let code_address = match self.env.tx.transact_to {
+            TransactTo::Call(code_address) => code_address,
+            TransactTo::Create => unimplemented!(), // TODO: implement creation
+        };
+        let bytecode = self
+            .db
+            .code_by_address(code_address)
+            .expect("failed to load bytecode");
+        let program = Program::from_bytecode(bytecode.as_slice()).unwrap();
```
MegaRedHand, line 60: Please add a TODO here and open an issue. We should map invalid/unknown opcodes to the INVALID operation.

src/program.rs
```diff
@@ -408,7 +408,7 @@ impl Operation {
             Operation::Revert => vec![Opcode::REVERT as u8],
             Operation::Mstore => vec![Opcode::MSTORE as u8],
             Operation::Mstore8 => vec![Opcode::MSTORE8 as u8],
-            Operation::Log(n) => vec![Opcode::LOG0 as u8 + n - 1],
+            Operation::Log(n) => vec![Opcode::LOG0 as u8 + n],
```
MegaRedHand, line 411: Good catch!

tests/ef_tests.rs
TomasArrachea: Done!

src/lib.rs
TomasArrachea: Done!

src/db.rs
MegaRedHand: The problem is `B256` is an alias for `U256`. This is wrong, since `B256` should be equivalent to `H256`.

MegaRedHand (approved):
LGTM

PR #268, opened 2024-06-14
Opcode CHAINID implementation

### LINKED ISSUE #52: Opcode: CHAINID

JulianVentura:
Closes #52 
This PR implements the CHAINID opcode with its tests.


ricomateo (approved):
LGTM

src/syscall.rs
```diff
@@ -187,6 +187,11 @@ impl<'c> SyscallContext<'c> {
         value.hi = (aux >> 128).low_u128();
     }
 
+    #[allow(improper_ctypes)]
```
azteca1998, line 190: Why is this necessary here and not in, for example, `get_calldata_size`?

src/syscall.rs
```diff
@@ -345,6 +351,10 @@ pub fn register_syscalls(engine: &ExecutionEngine) {
             symbols::STORE_IN_CALLVALUE_PTR,
             SyscallContext::store_in_callvalue_ptr as *const fn(*mut c_void, *mut U256) as *mut (),
         );
+        engine.register_symbol(
+            symbols::GET_CHAINID,
+            SyscallContext::get_chainid as *const fn(*mut c_void) as *mut (),
```
azteca1998, line 356: It shouldn't matter in this case as we're converting to a void pointer, but to be 100% correct we should specify the complete function's signature:
```
SyscallContext::get_chainid as *const extern "C" fn(&SyscallContext) -> u64 as *mut ()
```

azteca1998 (approved):

src/syscall.rs
JulianVentura: It seems it's not necessary, actually. I'll delete it.

src/syscall.rs
JulianVentura: Ok

PR #275, opened 2024-06-18
feat: implement BLOBHASH opcode

### LINKED ISSUE #55: Opcode: BLOBHASH

ricomateo:
Closes #55 
This PR implements code generation for the [BLOBHASH](https://www.evm.codes/#49?fork=cancun) opcode.


src/syscall.rs
```diff
@@ -311,6 +311,21 @@ impl<'c> SyscallContext<'c> {
         let log = Log { data, topics };
         self.inner_context.logs.push(log);
     }
+
+    pub extern "C" fn get_blob_hash_at_index(&mut self, index: &U256, blobhash: &mut U256) {
+        // supposing the index is a usize at most
+        // TODO: check if this is ok
+        let index = index.lo as usize;
+
+        if index < self.env.tx.blob_hashes.len() {
```
JulianVentura, line 320: Here you can remove the len check and call `self.env.tx.blob_hashes.get(index)` instead. That will return you an Option which you can match over.

JulianVentura (approved):
LGTM

ricomateo:
@MegaRedHand this one is ready for review

src/syscall.rs
```diff
@@ -311,6 +311,23 @@ impl<'c> SyscallContext<'c> {
         let log = Log { data, topics };
         self.inner_context.logs.push(log);
     }
+
+    pub extern "C" fn get_blob_hash_at_index(&mut self, index: &U256, blobhash: &mut U256) {
+        // supposing the index is a usize at most
+        // TODO: check if this is ok
+        let index = index.lo as usize;
+        let aux = self.env.tx.blob_hashes.get(index);
```
MegaRedHand, line 319: We can use `unwrap_or_default` here

src/syscall.rs
```diff
@@ -311,6 +311,23 @@ impl<'c> SyscallContext<'c> {
         let log = Log { data, topics };
         self.inner_context.logs.push(log);
     }
+
+    pub extern "C" fn get_blob_hash_at_index(&mut self, index: &U256, blobhash: &mut U256) {
+        // supposing the index is a usize at most
+        // TODO: check if this is ok
+        let index = index.lo as usize;
```
MegaRedHand, line 318: We should return zero when index is bigger than a `usize`

MegaRedHand (changes requested):

src/syscall.rs
```diff
@@ -451,6 +451,17 @@ impl<'c> SyscallContext<'c> {
         }
     }
 
+    pub extern "C" fn get_blob_hash_at_index(&mut self, index: &U256, blobhash: &mut U256) {
+        *blobhash = usize::try_from(index.lo).map_or(U256::default(), |idx| {
```
MegaRedHand, line 455: We should check `index.hi` is zero.

src/syscall.rs
```diff
@@ -451,6 +451,17 @@ impl<'c> SyscallContext<'c> {
...
+            self.env
+                .tx
+                .blob_hashes
+                .get(idx)
+                .cloned()
+                .map_or(U256::default(), |x| U256::from_be_bytes(x.into()))
+        });
```
MegaRedHand, line 462: This is a bit confusing. Generally, we have a chain of transformations (`map`, `and`, etc.), and at the end a single `unwrap`. In this case:

```suggestion
        *blobhash = usize::try_from(index.lo).and_then(|idx| {
            self.env
                .tx
                .blob_hashes
                .get(idx)
        })
        .map(U256::from_be_bytes)
        .unwrap_or_default();
```

MegaRedHand (changes requested):
Left some comments

src/syscall.rs
JulianVentura: Yep, that's way better. Thanks for the idea.

src/env.rs
```diff
@@ -1,5 +1,5 @@
 use crate::primitives::{Address, Bytes, U256};
-
+type B256 = U256;
```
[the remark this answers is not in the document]
JulianVentura: Done

src/syscall.rs
JulianVentura: Done

MegaRedHand (approved):
LGTM

PR #291, opened 2024-06-19
Implement opcode SELFBALANCE

### LINKED ISSUE #53: Opcode: SELFBALANCE

JulianVentura:
Closes #53 
This PR implements the opcode SELFBALANCE with tests.


FrancoGiachetta:
LGTM

FrancoGiachetta (approved):
LGTM

src/db.rs
```diff
@@ -37,17 +37,39 @@ impl Db {
...
+        contracts.insert(hash, bytecode);
+
+        Self {
+            accounts,
+            contracts,
+            block_hashes,
+        }
+    }
+
+    pub fn with_accounts(self, new_accounts: HashMap<Address, DbAccount>) -> Self {
```
MegaRedHand, line 62: Can we change this to receive a single account? Then we can just chain it when multiple accounts are needed.

src/db.rs
```diff
@@ -37,17 +37,39 @@ impl Db {
...
             bytecode_hash: hash,
             ..Default::default()
         };
-        db.accounts.insert(address, account);
-        db.contracts.insert(hash, bytecode);
-        db
+
+        let mut contracts = self.contracts.clone();
+        let mut accounts = self.accounts.clone();
+        let block_hashes = self.block_hashes;
```
MegaRedHand, line 50: No need to clone here. You can just declare `mut self`, modify that, and then return it.

src/db.rs
```diff
@@ -37,17 +37,39 @@ impl Db {
...
+            accounts,
+            contracts,
+            block_hashes,
+        }
+    }
+
+    pub fn with_accounts(self, new_accounts: HashMap<Address, DbAccount>) -> Self {
+        let contracts = self.contracts;
+        let block_hashes = self.block_hashes;
+        let mut accounts = self.accounts.clone();
```
MegaRedHand, line 65: Same here. Just declare `mut self`:

```rust
fn with_accounts(mut self, ...) -> Self {
    self.accounts.extend(...);
    self
}
```

src/syscall.rs
```diff
@@ -193,6 +193,20 @@ impl<'c> SyscallContext<'c> {
...
+        match self.db.basic(address).unwrap() {
+            Some(a) => {
+                balance.hi = (a.balance >> 128).low_u128();
+                balance.lo = a.balance.low_u128();
+            }
+            None => {
+                balance.hi = 0;
+                balance.lo = 0;
+            }
+        };
```
MegaRedHand, line 207: You can use `unwrap_or_default` to simplify this.

tests/evm.rs
```diff
@@ -1244,3 +1219,61 @@ fn balance_static_gas_check() {
...
+
+#[test]
+fn selfbalance_gas_check() {
+    let operations = vec![Operation::SelfBalance];
+    let mut env = Env::default();
+    env.tx.gas_limit = 999_999;
+    let needed_gas = gas_cost::SELFBALANCE;
+
+    run_program_assert_gas_exact(operations, env, needed_gas as _);
+}
```
MegaRedHand, line 1279: Can we add an example that checks that `SelfBalance` does the same as [`Address`, `Balance`] ?

MegaRedHand (changes requested):

tests/evm.rs
JulianVentura: Yes, that's a good one!

src/db.rs
```diff
@@ -36,18 +36,23 @@ impl Db {
...
-        db
+
+        self.accounts.insert(address, account);
+        self.contracts.insert(hash, bytecode);
+
+        Self {
+            accounts: self.accounts,
+            contracts: self.contracts,
+            block_hashes: self.block_hashes,
+        }
```
MegaRedHand, line 55:
```suggestion
        self
```

MegaRedHand (approved):
LGTM

PR #303, opened 2024-06-26
Implement CALL opcode

### LINKED ISSUE #76: Opcode: CALL

JulianVentura:
Closes #76 
This PR implements the CALL opcode with tests.


src/syscall.rs
```diff
@@ -50,6 +54,20 @@ impl U256 {
...
+        let lo_bytes = self.lo.to_be_bytes();
+        let address = [&hi_bytes[12..16], &lo_bytes[..]].concat();
+        Address::from_slice(&address)
+    }
+
+    pub fn to_primitive_u256(&self) -> EU256 {
+        let hi_bytes = self.hi.to_be_bytes();
+        let lo_bytes = self.lo.to_be_bytes();
+        let bytes = [&hi_bytes[..], &lo_bytes[..]].concat();
+        EU256::from_big_endian(&bytes)
```
MegaRedHand, line 69: Can we use shifting and adding instead of dealing with strings?

src/constants.rs
```diff
@@ -103,6 +104,20 @@ pub mod gas_cost {
     }
 }
 
+pub mod call_opcode {
+    // Return codes
+    pub const SUCCESS_RETURN_CODE: u32 = 1;
+    pub const REVERT_RETURN_CODE: u32 = 0;
+    pub const HALT_RETURN_CODE: u32 = 2;
```
MegaRedHand, line 111: These should be `u8`

src/constants.rs
```diff
@@ -103,6 +104,20 @@ pub mod gas_cost {
...
+
+    // Gas related constants
+    pub const WARM_MEMORY_ACCESS_COST: u64 = 100;
+    pub const NOT_ZERO_VALUE_COST: u64 = 9000;
+    pub const EMPTY_CALLE_COST: u64 = 25000;
```
MegaRedHand, line 116: typo
```suggestion
    pub const EMPTY_CALLEE_COST: u64 = 25000;
```

src/codegen/context.rs
```diff
@@ -887,4 +890,117 @@ impl<'c> OperationCtx<'c> {
...
+                location,
+                LoadStoreOptions::default(),
+            ))
+            .result(0)?
+            .into();
+        let gas_flag = consume_gas_as_value(context, start_block, consumed_gas)?;
+
+        // Compare the result value to see if its an error
+        // return_value == return_error_code -> HALT
+        let return_error_code = start_block
```
MegaRedHand, line 962: We don't necessarily need to halt here right? HALT is the same as REVERTED in spec terms, and when the inner context reverts a 0 is pushed to the stack while execution continues.

tests/evm.rs
```diff
@@ -1359,3 +1373,297 @@ fn gaslimit_stack_overflow() {
     let (env, db) = default_env_and_db_setup(program);
     run_program_assert_halt(env, db);
 }
+
+#[test]
+fn call_returns_addition_from_arguments() {
+    let (a, b) = (BigUint::from(3_u8), BigUint::from(5_u8));
+    let db = Db::new();
+
+    // Calee
```
MegaRedHand, line 1382: typo
```suggestion
    // Callee
```

tests/evm.rs
```diff
@@ -55,6 +55,20 @@ fn run_program_assert_halt(env: Env, db: Db) {
     assert!(result.is_halt());
 }
 
+fn run_program_assert_gas_exact_aux(mut env: Env, db: Db, needed_gas: u64) {
```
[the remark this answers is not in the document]
JulianVentura: I have used that not-so-nice name on purpose. I created the issue #310 to modify the name of it and refactor the already existing tests which are using the old function with similar name `run_program_assert_gas_exact`. Anyways, I can change the name and modify that issue if you want.

src/syscall.rs
JulianVentura: Is `concat` method using strigns? If I'm not mistaken, [this is the documentation](https://doc.rust-lang.org/1.30.0/std/slice/trait.SliceConcatExt.html), and it doesn't seem to be using strings underneath. I can implement this using shift and additions as you mentioned, but it will still need two `U256` allocs. We would have to run some benchmarks in order to decide which is faster. Do you want me to make the change?

src/codegen/context.rs
JulianVentura: Ok, I didn't know that HALT was the same as REVERT in spec terms. I'll check into that.

tests/evm.rs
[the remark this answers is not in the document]
MegaRedHand: Ok, but please change the name. I don't think we need to move all tests to use this function, but we may refactor the other one to call this one.

src/syscall.rs
MegaRedHand: With strings, I'm referring to byte strings (i.e. slices or `&[u8]`) not `String`s. I'm asking for this change because dealing only with numbers makes code simpler to understand. Moving on to bytes makes us have to deal with things like endianness.

src/constants.rs
JulianVentura: Done

src/codegen/context.rs
JulianVentura: Done

src/constants.rs
JulianVentura: Done

tests/evm.rs
JulianVentura: Done

src/syscall.rs
JulianVentura: Done

src/codegen/operations.rs
```diff
@@ -4568,3 +4570,103 @@ fn codegen_gaslimit<'c, 'r>(
...
+        .result(0)?
+        .into();
+    let ret_size = ok_block
+        .append_operation(arith::trunci(ret_size, uint32.into(), location))
+        .result(0)?
+        .into();
+
+    let args_block = region.append_block(Block::new(&[]));
+    let ret_block = region.append_block(Block::new(&[]));
+    expand_memory_from_offset_and_size(
```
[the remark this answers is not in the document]
JulianVentura: Done

tests/evm.rs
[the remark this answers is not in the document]
JulianVentura: Done

src/codegen/context.rs
```diff
@@ -966,4 +969,90 @@ impl<'c> OperationCtx<'c> {
...
+            context,
+            gas_flag,
+            finish_block,
+            &self.revert_block,
+            &[],
+            &[],
+            location,
+        ));
+
+        // Extend the 32 bits result to 256 bits
```
MegaRedHand, line 1048:
```suggestion
        // Extend the 8 bits result to 256 bits
```

src/db.rs
```diff
@@ -110,6 +110,14 @@ pub struct AccountInfo {
     pub code: Option<Bytecode>,
 }
 
+impl AccountInfo {
+    pub fn is_empty(&self) -> bool {
+        self.balance.is_zero()
+            && self.nonce == 0
+            && self.code.clone().unwrap_or_default().is_empty()
```
MegaRedHand, line 117: Instead of cloning here, we can check if the `code_hash` equals the empty hash: `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`. Besides, `code` is just the cached value. `None` only means "this should be fetched from DB".

src/syscall.rs
```diff
@@ -220,6 +238,141 @@ impl<'c> SyscallContext<'c> {
...
+        env.tx.gas_limit = gas_to_send;
+
+        //Copy the calldata from memory
+        let off = args_offset as usize;
+        let size = args_size as usize;
+        env.tx.data = Bytes::from(self.inner_context.memory[off..off + size].to_vec());
+
+        //NOTE: We could optimize this by not making the call if the bytecode is zero.
+        //We would have to refund the stipend here
+        let bytecode = self.db.code_by_address(callee_address);
```
MegaRedHand, line 330: This should use `db.basic` -(if code is None)-> `db.code_by_hash`. In fact, we could add it to the `Database` trait with a default implementation.

tests/evm.rs
```diff
@@ -1833,3 +1847,315 @@ fn blobhash_with_index_too_big() {
...
+        Operation::Push((1_u8, BigUint::from(args_size))), //Args size
+        Operation::Push((1_u8, BigUint::from(args_offset))), //Args offset
+        Operation::Push((1_u8, BigUint::from(value))),    //Value
+        Operation::Push((16_u8, BigUint::from_bytes_be(callee_address.as_bytes()))), //Address
+        Operation::Push((1_u8, BigUint::from(gas))),      //Gas
+        Operation::Call,
+    ];
+
+    //address_access_cost + positive_value_cost + value_to_empty_account_cost
+    let caller_call_cost = 100 + 9000 + 25000;
```
MegaRedHand, line 2148: Can we use the constants here?

src/syscall.rs
```diff
@@ -220,6 +238,141 @@ impl<'c> SyscallContext<'c> {
...
+        *consumed_gas += gas_to_send;
+        gas_to_send += stipend;
+
+        let mut env = self.env.clone();
+        env.tx.transact_to = TransactTo::Call(callee_address);
+        //TODO: Check if this is ok
+        env.tx.caller = match self.env.tx.transact_to {
+            TransactTo::Call(a) => a,
+            TransactTo::Create => Address::zero(),
+        };
```
MegaRedHand, line 319: This shouldn't be changed, as the `ORIGIN` remains the same. The `CALLER` should be taken from somewhere else. Maybe some struct representing the "call frame", which stores related metadata.

MegaRedHand (changes requested):

src/codegen/context.rs
JulianVentura: Done

src/db.rs
JulianVentura: Done

tests/evm.rs
JulianVentura: Done

src/syscall.rs
JulianVentura: Done

MegaRedHand (approved):
Good work!

src/syscall.rs
JulianVentura: Done

src/db.rs
```diff
@@ -146,6 +134,9 @@ pub trait Database {
 
     /// Get block hash by block number.
     fn block_hash(&mut self, number: U256) -> Result<B256, Self::Error>;
+
+    /// Get account code by its address.
+    fn code_by_address(&mut self, address: Address) -> Result<Bytecode, Self::Error>;
```
MegaRedHand, line 146: Let's add a default implementation so it remains backwards-compatible

src/syscall.rs
```diff
@@ -653,8 +657,9 @@ impl<'c> SyscallContext<'c> {
     }
 
     pub extern "C" fn get_codesize_from_address(&mut self, address: &U256) -> u64 {
+        //TODO: Here we are returning 0 if a Database error occurs. Check this
```
MegaRedHand, line 673: Let's add a `eprintln!` call on fail (and leave the TODO).

src/syscall.rs
```diff
@@ -734,7 +739,9 @@ impl<'c> SyscallContext<'c> {
             self.inner_context.memory[dest_offset..dest_offset + size].fill(0);
             return;
         };
-        let code = self.db.code_by_address(address);
+        // TODO: Check if returning default bytecode on database failure is ok
+        // A silenced error like this may produce unexpected code behaviour
+        let code = self.db.code_by_address(address).unwrap_or_default();
```
MegaRedHand, line 757: Same

MegaRedHand (changes requested):

src/db.rs
JulianVentura: Done

src/syscall.rs
JulianVentura: Done

src/syscall.rs
JulianVentura: Done

MegaRedHand (approved):

PR #304, opened 2024-06-27
add gas consumption to SSTORE operation

### LINKED ISSUE #283: Compute dynamic gas cost for SSTORE opcode

TomasArrachea:
Closes #283 

This PR implements the dynamic gas costs of the SSTORE operation.


src/syscall.rs
```diff
@@ -337,13 +343,94 @@ impl<'c> SyscallContext<'c> {
...
+                gas_cost = 20_000;
+            } else {
+                gas_cost = 2_900;
+            }
+        } else {
+            gas_cost = 100;
+        }
+        if is_cold {
+            gas_cost += 2_100;
+        }
```
MegaRedHand, line 396: Can you simplify this?

src/syscall.rs
```diff
@@ -337,13 +343,94 @@ impl<'c> SyscallContext<'c> {
...
+                }
+                if value == original {
+                    if original.is_zero() {
+                        gas_refund += 19_900;
+                    } else {
+                        gas_refund += 2_800;
+                    }
+                }
+            }
+        }
```
MegaRedHand, line 421: This too.

src/codegen/operations.rs
```diff
@@ -2324,9 +2325,57 @@ fn codegen_sstore<'c, 'r>(
...
+        .append_operation(
+            ods::llvm::icmp(
+                context,
+                IntegerType::new(context, 1).into(),
+                zero_constant_value,
+                write_result,
+                IntegerAttribute::new(uint64.into(), 0).into(),
+                location,
+            )
+            .into(),
```
MegaRedHand, line 2362: We can use `arith::cmpi` that has simpler parameters.

src/codegen/operations.rs
```diff
@@ -2324,9 +2325,57 @@ fn codegen_sstore<'c, 'r>(
...
+        )
+        .result(0)?
+        .into();
+    let return_block = region.append_block(Block::new(&[]));
```
MegaRedHand, line 2366: Let's rename it to `end_block`.

src/syscall.rs
```diff
@@ -102,8 +105,10 @@ pub struct InnerContext {
     // The program bytecode
     pub program: Vec<u8>,
     gas_remaining: Option<u64>,
+    gas_refund: Option<i64>,
```
MegaRedHand, line 108: 1. Why an `Option`?
2. Why signed?

src/syscall.rs
```diff
@@ -337,13 +343,94 @@ impl<'c> SyscallContext<'c> {
...
+        // Check if gas is enough, and then perform the write operation
+        // The amount of gas left to the transaction hass to be less than or equal 2300 (since Istanbul fork).
+        if gas_cost + (gas_cost::SSTORE_MIN_REMAINING_GAS as u64) > *gas_remaining {
+            return 1;
+        }
+        *gas_remaining -= gas_cost;
+        let current_refund = self.inner_context.gas_refund.unwrap_or(0);
+        self.inner_context.gas_refund = Some(gas_refund + current_refund);
+
+        self.db.write_storage(self.env.tx.caller, key, value);
```
MegaRedHand, line 432: We should do this only after the transaction finishes.

src/syscall.rs
```diff
@@ -337,13 +343,94 @@ impl<'c> SyscallContext<'c> {
...
+                    }
+                }
+            }
+        }
+
+        // Check if gas is enough, and then perform the write operation
+        // The amount of gas left to the transaction hass to be less than or equal 2300 (since Istanbul fork).
+        if gas_cost + (gas_cost::SSTORE_MIN_REMAINING_GAS as u64) > *gas_remaining {
+            return 1;
+        }
```
MegaRedHand, line 427: Can we return the gas cost instead and check this in the Rust side?

src/codegen/operations.rs
```diff
@@ -2282,51 +2281,86 @@ fn codegen_sstore<'c, 'r>(
...
+            IntegerAttribute::new(uint64, gas_cost::SSTORE_MIN_REMAINING_GAS).into(),
             location,
-            AllocaOptions::new().elem_type(Some(TypeAttribute::new(uint256.into()))),
         ))
         .result(0)?
         .into();
+    let needed_gas = ok_block
+        .append_operation(arith::addi(min_remaining_gas, gas_cost, location))
+        .result(0)?
+        .into();
```
MegaRedHand, line 2301: Let's do `gas_counter - gas_cost` first, so we avoid adding here.

src/syscall.rs
```diff
@@ -331,19 +341,96 @@ impl<'c> SyscallContext<'c> {
...
         let key = u256_from_u128(stg_key.hi, stg_key.lo);
 
-        let result = self.db.read_storage(address, key);
+        // Read value from journaled_storage. If there isn't one, then read from db
+        let result = self
+            .inner_context
+            .journaled_storage
+            .get(&key)
+            .map(|slot| slot.present_value)
+            .unwrap_or(self.db.read_storage(address, key));
```
MegaRedHand, line 350: We use `*_or_else` functions when computation should be done lazily (like here).
```suggestion
            .unwrap_or_else(|| self.db.read_storage(address, key));
```

MegaRedHand (changes requested):
Looks good! Left one coment, but should be good to go once fixed

src/syscall.rs
```diff
@@ -384,46 +384,40 @@ impl<'c> SyscallContext<'c> {
...
-        if value == current {
-            gas_cost = 100;
-        } else if current == original {
-            if original.is_zero() {
-                gas_cost = 20_000;
-            } else {
-                gas_cost = 2_900;
-            }
+        let mut gas_cost: i64 = if original.is_zero() && current.is_zero() {
+            20_000
```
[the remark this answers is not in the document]
TomasArrachea: You are right, but i think that condition is missing the case where `original == current == value`. I made some changes

MegaRedHand (approved):
LGTM

PR #316, opened 2024-07-02
Increase benchmarks precision

JulianVentura:
This PR modifies the benchmarks precision by incrementing `hyperfine` rounds from `10` to `20`.
It also disables memory swap, in order to reduce noise.


.github/scripts/run-benchmarks.sh
```diff
@@ -2,9 +2,10 @@
 
 
 echo "# Benchmarking results" > bench-hyperfine.md
+sudo swapoff -a # Disabling swap memory to reduce noice
```
MegaRedHand, line 5:
```suggestion
sudo swapoff -a # Disabling swap memory to reduce noise
```

MegaRedHand:
The changes weren't effective

PR #318, opened 2024-07-02
Run ef tests

### LINKED ISSUE #311: Update EF test runner
Load EF tests and parse it to run with our public API

TomasArrachea:
Closes #311 .

This PR updates the EF tests runner to load the parameters using our public API and check the results.
Also, refactored opcode BLOBBASEFEE to set blob gas price from parameter `excess_blob_gas`.

Found bugs:
- PUSH bytecode parsing (fixed)
- CODECOPY array accesing (left as TODO) 

Running the following test groups (previously ignored):
- `stEIP3855-push0`
- `vmBitwiseLogicOperation`
- `eip1344_chainid`
- `vmIOandFlowOperations`
- `stRecursiveCreate`
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


src/db.rs
```diff
@@ -29,10 +29,17 @@ impl Db {
         Self::default()
     }
 
-    pub fn update_account(&mut self, address: Address, nonce: u64, balance: U256) {
+    pub fn update_account(
```
MegaRedHand, line 32: Let's call it `set_account` instead
```suggestion
    pub fn set_account(
```

src/utils.rs
```diff
@@ -1558,6 +1559,49 @@ pub(crate) fn get_basefee<'a>(
...
+///
+/// This is used to calculate the blob price.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers)
+/// (`fake_exponential`).
+///
+/// # Panics
+///
+/// This function panics if `denominator` is zero.
+#[inline]
```
MegaRedHand, line 1585: Let's remove this

src/utils.rs
```diff
@@ -1558,6 +1559,49 @@ pub(crate) fn get_basefee<'a>(
     Ok(basefee)
 }
 
+/// Calculates the blob gas price from the header's excess blob gas field.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers)
+/// (`get_blob_gasprice`).
+#[inline]
```
MegaRedHand, line 1566: This too

src/syscall.rs
```diff
@@ -321,7 +319,13 @@ impl<'c> SyscallContext<'c> {
             size
         };
 
-        let code_slice = &self.inner_context.program[code_offset..code_offset + size];
+        let Some(code_slice) = &self
+            .inner_context
+            .program
+            .get(code_offset..code_offset + size)
+        else {
+            return; // TODO: fix bug with code indexes
```
MegaRedHand, line 327: Let's add a call to `error!`, just for good measure

tests/ef_tests.rs
```diff
@@ -128,23 +127,93 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
         };
-        let mut env = Env::default();
-        env.tx.transact_to = TransactTo::Call(to);
+        let sender = unit.transaction.sender.unwrap_or_default();
+        let gas_price = unit.transaction.gas_price.unwrap_or_default();
+
+        // NOTE: currently we only support Cancun spec
+        let Some(tests) = unit.post.get("Cancun") else {
+            return Ok(());
+        };
```
MegaRedHand, line 146: Let's move this to the beginning of the loop, and use `continue` instead of `return`.

tests/ef_tests.rs
```diff
@@ -128,23 +127,93 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
+                    account_info.nonce,
+                    account_info.balance,
+                    account_info.storage.clone(),
+                );
+            }
+            let mut evm = Evm::new(env, db);
+
+            let res = evm.transact().unwrap();
+
+            if test.expect_exception.is_some() {
```
MegaRedHand, line 186: Shouldn't we check the expected error message?

TomasArrachea: `expect_exception` only represents an error description, so we can't really compare it to the returned error

MegaRedHand (approved):
Good work! 🚀

PR #320, opened 2024-07-03
Allow to optionally set the output files paths

### LINKED ISSUE #319: Allow specifying (or not) a path for dumping intermediate artifacts
We can use something similar to Concrete's `Session`:

https://github.com/lambdaclass/concrete/blob/bc72356bb2116805dc737c1ea2edd0a045249e77/crates/concrete_session/src/lib.rs#L11-L17

JulianVentura:
### Inspiration

Currently, we are always writing both mlir after-pass and output files to disk. This is probably introducing noise in our benchmarks running on GitHub, which reduces precision by a big margin. This PR adds a new `Session` struct which allows to set up if and where output and intermediate files should be stored.

### Changes

* Added `Session` struct in `src/context.rs`
* Modify `compile()` function from `src/context.rs` to receive a `Session` instead of a output file path.
* Modify calls to `compile()` function to cope with new API.
  * Eliminate output file writing from benchmarks

Closes #319

### DIFF
--- a/tests/operations.rs
+++ b/tests/operations.rs
@@ -4,7 +4,7 @@
 //! may not work properly.
 use evm_mlir::{
     constants::gas_cost::{self, log_dynamic_gas_cost},
-    context::Context,
+    context::{Context, ContextConfig},
     db::Db,
     env::Env,
     executor::Executor,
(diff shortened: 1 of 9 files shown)


tests/operations.rs
```diff
@@ -16,21 +16,17 @@ use evm_mlir::{
 use hex_literal::hex;
 use num_bigint::{BigInt, BigUint};
 use rstest::rstest;
-use tempfile::NamedTempFile;
```
MegaRedHand, line 19: Can we remove the `tempfile` dependency?

JulianVentura: Done

PR #324, opened 2024-07-04
feat: implement CREATE opcode

### LINKED ISSUE #75: Opcode: CREATE

TomasArrachea:
Closes #75 

This PR implements the code generation for the [CREATE](https://www.evm.codes/#f0?fork=cancun) opcode


src/db.rs
```diff
@@ -50,18 +50,23 @@ impl Db {
     }
 
     pub fn with_bytecode(mut self, address: Address, bytecode: Bytecode) -> Self {
```
MegaRedHand, line 52: Let's rename this to `with_contract`.

src/db.rs
```diff
@@ -50,18 +50,23 @@ impl Db {
     }
```
MegaRedHand, line 50: Unrelated, but `set_account` should initialize empty accounts.

src/syscall.rs
```diff
@@ -850,9 +855,79 @@ impl<'c> SyscallContext<'c> {
...
-        address.lo = u128::from_be_bytes(lo.try_into().unwrap());
-        address.hi = u128::from_be_bytes(hi.try_into().unwrap());
+        *address = U256::from_fixed_be_bytes(hash.to_fixed_bytes());
+    }
+
+    pub extern "C" fn create(&mut self, size: u32, offset: u32, value: &mut U256) {
+        let value_as_u256 = value.to_primitive_u256();
+        let offset = offset as usize;
+        let size = size as usize;
+        let sender_address = self.env.tx.get_address();
```
MegaRedHand, line 865: This assumes the caller is the contract that first received the transaction. Here we should retrieve the caller from the stack frame.

src/syscall.rs
```diff
@@ -850,9 +855,79 @@ impl<'c> SyscallContext<'c> {
...
+        let encoded_nonce = encode_rlp_u64(sender_account.nonce);
+        let mut buf = Vec::<u8>::new();
+        buf.push(0xd5);
+        buf.extend_from_slice(&encoded_nonce.len().to_be_bytes());
+        buf.push(0x94);
+        buf.extend_from_slice(sender_address.as_bytes());
+        buf.extend_from_slice(&encoded_nonce);
+        let mut hasher = Keccak256::new();
+        hasher.update(&buf);
+        let dest_addr = Address::from_slice(&hasher.finalize()[12..]);
```
MegaRedHand, line 886: Can we put this inside a helper function?

src/syscall.rs
```diff
@@ -850,9 +855,79 @@ impl<'c> SyscallContext<'c> {
...
+
+        // Check if there is already a contract stored in dest_address
+        if let Ok(Some(_)) = self.db.basic(dest_addr) {
+            *value = U256::zero();
+            return;
+        }
```
MegaRedHand, line 892: We must revert if a contract already exists.

tests/evm.rs
```diff
@@ -2777,3 +2778,149 @@ fn returndatacopy_gas_check() {
...
+    let mut evm = Evm::new(env, db);
+    let result = evm.transact().unwrap().result;
+    assert!(result.is_success());
+
+    // Check that contract is created correctly in the returned address
+    let returned_addr = Address::from_slice(&result.output().unwrap()[12..]);
+    let new_account = evm.db.basic(returned_addr).unwrap().unwrap();
+    assert_eq!(new_account.balance, EU256::from(value));
+    assert_eq!(new_account.nonce, 1);
+
```
MegaRedHand, line 2823: Shouldn't we check the contract's code?

MegaRedHand (changes requested):

src/syscall.rs
TomasArrachea: The env.tx.to address actually stores the callee address, that represents the address of the contract being executed here, and that address acts as the sender of the create

MegaRedHand (approved):
LGTM

PR #327, opened 2024-07-05
Fix on `CALL` opcode memory copying

### LINKED ISSUE #325: Fix `CALL` opcode memory copy bug
Currently, `CALL` opcode is using `Vec<u8>::copy_from_slice`  method when copying callee return data into memory on `src/syscall.rs`. This method will fail if the returned data size is different from the specified (by argument) `ret_size`. We should add some checks and tests in order to act accordingly to the specs for this opcode, which can be tested on [EVM Codes playground](https://www.evm.codes/playground?fork=cancun&unit=Wei&codeType=Mnemonic&code='W23cx6d64333333333g5601bfge6012f3~_k23qsizek9qoffset~0qvaluejCREATEj~32cxQQQQk_%20jjY%20Call%20with%20no%20parameters%2C%20returncXSizeXOffsetVVVjDUP6~2cxZjCALL'~jWyZZq%20Yk~1%20j%5Cng3600052600c%200_0jMSTOREZFFFFY%2F%2FXVqretWPUSHVk0Qyy%01QVWXYZ_cgjkqy~_)

JulianVentura:
### Inspiration

This PR fixes memory copy bugs on `CALL` opcode, as explained in #325. 

### Changes

* Added `copy_exact()` function on `SyscallContext`
* Deleted function `write_memory` on `SyscallContext`
* Modified call copy memory logic on `SyscallContext::call`
* Added tests to validate changes

Closes #325

### DIFF
--- a/src/syscall.rs
+++ b/src/syscall.rs
@@ -367,7 +367,12 @@ impl<'c> SyscallContext<'c> {
             ExecutionResult::Success {
                 gas_used, output, ..
             } => {
-                self.write_to_memory(ret_offset as _, ret_size as _, output.data());
+                Self::copy_exact(
+                    &mut self.inner_context.memory,
+                    output.data(),
+                    ret_offset,
+                    ret_size,
+                );
                 *consumed_gas -= gas_to_send - gas_used;
                 call_opcode::SUCCESS_RETURN_CODE
             }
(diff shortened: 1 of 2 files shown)


MegaRedHand (approved):
LGTM

PR #331, opened 2024-07-10
Implement opcode `STATICCALL`

JulianVentura:
This PR implements the opcode `STATICCALL` with tests.

Closes #81

PR #352, opened 2024-07-12
Fix on `CALLER` opcode syscall

### LINKED ISSUE #292: Change CALLER opcode implementation to return proper value
For the moment we have implemented `CALLER` opcode returning the `tx.env.caller` value from the `struct Env`. This is only correct if we are not making any call to another contract. We should change this in the future when we implement any of the call opcodes, for example `CALL`.

JulianVentura:
### Description

Currently, opcode `CALLER` is behaving like the `ORIGIN` opcode, returning the `env.tx.caller` instead of `CallFrame::caller`.
This PR fixes this behavior.

### Changes

* Updated `SyscallContext::store_in_caller_ptr` on `src/syscall.rs`.
* Updated `CALL` and `STATICCALL` tests to test this functionality

Depends on #331 
Closes #292

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ An EVM-bytecode to machine-bytecode compiler using MLIR and LLVM.
 1. (0xF0) CREATE
 1. (0xF1) CALL
 1. (0xF3) RETURN
+1. (0xF5) CREATE2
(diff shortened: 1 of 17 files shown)


MegaRedHand (approved):
Good catch!

PR #364, opened 2024-07-20
fix: dead link on `./src/main.rs`

Peefy:


mpaulucci:
Sorry, we're not accepting these type of fixes atm

Issue #227, opened 2024-06-11
Add dynamic gas cost for SLOAD opcode


FrancoGiachetta:

FrancoGiachetta:
#60

PR #371, opened 2024-08-28
feat: pass ef_test stSLoadTest

belfortep:
Closes https://github.com/lambdaclass/evm_mlir/issues/370
Closes https://github.com/lambdaclass/evm_mlir/issues/227

This PR fixs the ef_test for stSLoadTest, adding the warm/cold gas cost


src/syscall.rs
```diff
@@ -1104,7 +1111,7 @@ impl<'c> SyscallContext<'c> {
             );
             engine.register_symbol(
                 symbols::STORAGE_READ,
-                SyscallContext::read_storage as *const fn(*const c_void, *const U256, *mut U256)
+                SyscallContext::read_storage as *const fn(*mut c_void, *const U256, *mut U256)
```
juanimedone, line 1114: Is this change needed?

belfortep: Oh you are right, the `self` value isn't modified, it could be `*const` and the syscall receive a `&self`

belfortep: Never mind, the `*mut` is needed because the syscall use the `read_storage` function that needs a mut reference, I think it works either way but maybe it could generate undefined behaviour with the `*const`

juanimedone: Ok yes, we should leave it with `*mut`.

juanimedone (approved):
Looks good to me!

MegaRedHand (approved):
Great work!!

Issue #351, opened 2024-07-11
Implement Access List


TomasArrachea:
Implementing the [ACCESS LIST](https://www.evm.codes/about#access_list) will allow to mark addresses as warm/cold and use it for gas calculation of several opcodes.

Issue #281, opened 2024-06-19
Compute dynamic gas cost for EXT* opcodes


ricomateo:
This issue involves handling cold and warm address accesses

PR #373, opened 2024-08-28
First implementation of access lists

### LINKED ISSUE #307: Add warm and cold address access to CALL opcode
Currently we are calculating the CALL opcode gas cost, but we are missing to account for cold or warm address access. We are calculating the account access cost as if every address were warm. This should be changed.

### LINKED ISSUE #323: Add warm and cold address access to EXTCODEHASH opcode

belfortep:
Closes https://github.com/lambdaclass/evm_mlir/issues/374
Closes https://github.com/lambdaclass/evm_mlir/issues/351
Closes https://github.com/lambdaclass/evm_mlir/issues/281
Closes #323 
Closes #307 
## Summary
This PR try to add `access lists` to the code, as it will be needed for different calculations of gas costs, also fix the EIP2930 tests
## Changes
- Added the `access lists` struct, with methods that matches the used in other implementations
- Added tests for different use cases of the `access lists` testing the reduced costs of gas
- The opcodes now modify the `access lists` as explained [here](https://www.evm.codes/about#access_list)


tests/evm.rs
```diff
@@ -4490,66 +4490,3 @@ fn keys_in_access_list_are_warm() {
     env.tx.access_list = access_list;
     run_program_assert_gas_exact(program, env, used_gas as _);
 }
-
-#[test]
-fn precompile_staticcall_with_access_list_is_warm() {
```
MegaRedHand: Why were these tests removed?

belfortep: It tested the same behaviour as the `addresses_in_access_list_are_warm`, just changing the address to a precompile, thought it was unnecesary

MegaRedHand: Ok. But these weren't the tests that failed, right?

belfortep: Made an oopsie, the test I wanted to delete was the `coinbase` one, this test was the failing one that used the staticcall (now I understand why the CI started working)

src/env.rs
```diff
@@ -1,17 +1,94 @@
...
+    /// Adds the precompile addresses to the access list as they are always accessed as warm.
+    pub fn add_precompile_addresses(&mut self) {
+        self.add_address(Address::from_low_u64_be(precompiles::BLAKE2F_ADDRESS));
+        self.add_address(Address::from_low_u64_be(precompiles::ECRECOVER_ADDRESS));
+        self.add_address(Address::from_low_u64_be(precompiles::IDENTITY_ADDRESS));
+        self.add_address(Address::from_low_u64_be(precompiles::MODEXP_ADDRESS));
+        self.add_address(Address::from_low_u64_be(precompiles::RIPEMD_160_ADDRESS));
+        self.add_address(Address::from_low_u64_be(precompiles::SHA2_256_ADDRESS));
+    }
+
```
lazcanoluca, line 68: Can you create a couple of issues so we don't forget to add the remaining precompiles here?

lazcanoluca (approved):
lgtm

src/env.rs
belfortep: Just created the issue🫡

src/syscall.rs
```diff
@@ -653,7 +665,8 @@ impl<'c> SyscallContext<'c> {
             return 0;
         };
 
-        let is_cold = !self.journal.key_is_warm(&address, &key);
+        let is_cold = !(self.journal.key_is_warm(&address, &key)
```
JulianVentura, line 668: Why are we asking both the journal and the access list for key warm/cold status? Can't we introduce all keys under `access_list` directly into journal cold/warm state from start?

JulianVentura (changes requested):

src/syscall.rs
belfortep: Oh, I had to ask that, maybe we can move the verification of warm keys from the `journal` directly to the `access_list`? If not, changing it to adding the keys to the `journal` from the start seems good

JulianVentura: Yeah both solutions would be ok. I actually like a bit more moving that responsability into Journal because I think it simplifies keys and addresses warm/cold state querying, and because it also simplifies how we may propagate those states back and forth from caller/callee on any external contract call. Journal was designed to wrap storage and key/address access state responsabilities with that in mind (and this of course has some drawbacks!).

We may change it in the future if we realize it's not a good idea.

belfortep: I'm facing a little problem with this, the Journal have the keys with the slots that have values associated to them, but in the AccessList, we only know the key we are using not what is in there, is it fine if we initialized that storage slot with default values?

JulianVentura (changes requested):

bench/revm_comparison/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub fn run_with_evm_mlir(program: &str, runs: usize, number_of_iterations: u32)
     calldata[28..32].copy_from_slice(&number_of_iterations.to_be_bytes());
     env.tx.data = Bytes::from(calldata);
     let mut db = Db::default();
-    let journal = Journal::new(&mut db);
+    let journal = Journal::new(&mut db, &env.tx.access_list);
```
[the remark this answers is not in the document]
belfortep: But creating the Journal and then adding the values couldn't lead to problems if we forgot them? If we are going to always add the accounts from the access lists to the journal it could make sense to add it to the ::new()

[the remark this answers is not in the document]
JulianVentura: It's actually a matter of preference. Both solutions are valid.
I like to separate access lists from the Journal constructor because you don't really need an access list to build a Journal, it's not a requirement. If for some reason we don't care passing an access list to the Journal, then we would have to pass an empty array or a `Default::default()` and that is a bit messy to me.
I think that a good solution would be to add a `with_prefetch(accounts: Vec<(Address, Vec<U256>)>)` method so you can have something like so:

```rust
let journal = Journal::new(&mut db).with_prefetch(&accounts);
```

This way you allow to instantiate the Journal without needing an access list and you keep the api clean so there is no code repetition.

src/env.rs
```diff
@@ -1,17 +1,94 @@
...
         },
-        MAX_BLOB_NUMBER_PER_BLOCK, VERSIONED_HASH_VERSION_KZG,
+        precompiles, MAX_BLOB_NUMBER_PER_BLOCK, VERSIONED_HASH_VERSION_KZG,
     },
     primitives::{Address, Bytes, B256, U256},
     result::InvalidTransaction,
     utils::calc_blob_gasprice,
 };
 
+#[derive(Clone, Debug, Default)]
```
JulianVentura, line 17: Do we really need to have this structure and functions?
We may need the gas calculation one, but I think it can be moved to some `utils` or something like that, so we keep the code base smaller and simpler.

src/syscall.rs
```diff
@@ -192,6 +192,7 @@ impl<'c> SyscallContext<'c> {
     pub fn get_result(&self) -> Result<ResultAndState, EVMError> {
         let gas_remaining = self.inner_context.gas_remaining.unwrap_or(0);
         let gas_initial = self.env.tx.gas_limit;
+        // TODO: Probably here we need to add the access_list_cost to gas_used, but we need a refactor of most tests
```
JulianVentura, line 195: We should create an issue for this.

src/syscall.rs
```diff
@@ -789,7 +789,9 @@ impl<'c> SyscallContext<'c> {
 
     pub extern "C" fn get_codesize_from_address(&mut self, address: &U256) -> u64 {
         //TODO: Here we are returning 0 if a Database error occurs. Check this
-        self.journal.code_by_address(&Address::from(address)).len() as _
+        let codesize = self.journal.code_by_address(&Address::from(address)).len();
```
JulianVentura, line 792: Aren't we missing a check here for address warm/cold state?

JulianVentura: We should call `account_is_warm` first, then `code_by_address`. 
I don't think we need an `add_account_as_warm` method if `code_by_address` already does a fetch (so it becomes warm). 

All of this should be tested.

src/syscall.rs
```diff
@@ -341,15 +344,13 @@ impl<'c> SyscallContext<'c> {
                 let call_type = CallType::try_from(call_type)
                     .expect("Error while parsing CallType on call syscall");
 
-                //TODO: This should instead add the account fetch (warm or cold) cost
-                //For the moment we consider warm access
                 let callee_account = match self.journal.get_account(&callee_address) {
```
JulianVentura, line 347: What happens if the account exists, but it is cold? I think you are handling it as warm here.

src/syscall.rs
```diff
@@ -839,7 +843,16 @@ impl<'c> SyscallContext<'c> {
                     balance.lo = 0;
                 }
             };
+
+            let is_cold = !self.journal.account_is_warm(&address);
```
JulianVentura, line 847: This will always return `warm`, because you are fetching that account before!

JulianVentura (commented):
Here are some more comments.
As a general note, I think there are some bugs that we are not seeing because we are missing some tests or we are not testing properly.

Give them a look, please.

bench/revm_comparison/src/lib.rs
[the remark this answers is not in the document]
belfortep: I really like that way with the prefetch, you are right of passing an empty array or default could be bothersome, also the Journal don't really need an Access List, only needs the element inside it

src/syscall.rs
belfortep: Ah, so we should make that `code_by_address` be the one that makes the account warm/cold here?

src/syscall.rs
belfortep: I'm getting a little dizzy, accounts in the journal are always `warm`, or we can have `cold` ones?

src/syscall.rs
JulianVentura: That's how we detect warm/cold state on an account if I'm not mistaken, by checking if it is cached in the journal internal `HashMap`.

src/syscall.rs
belfortep: > What happens if the account exists, but it is cold? I think you are handling it as warm here.

Okay, but in this other comment aren't we doing the same but we need to verify the warm/cold access? or I'm missing something?

src/syscall.rs
belfortep: Just added some tests and seems the `add_account_as_warm` is needed if the address used is not in the journal from the start

TomasArrachea (approved):
LGTM

src/codegen/operations.rs
```diff
@@ -128,6 +128,64 @@ pub fn generate_code_for_op<'c>(
     }
 }
 
+fn get_gas_ptr<'c>(
```
JulianVentura, line 131: This function doesn't actually return the gas ptr, does it? It allocates new memory where it stores a copy of the current gas counter. I would change the name to reflect that, because it may be misleading. 
And I would move this function to `src/utils.rs` with the other auxiliary functions.

JulianVentura (approved):
LGTM!
Left you a small comment. You can merge after solving it.

src/codegen/operations.rs
belfortep: You are right, changed it to `allocate_gas_counter_ptr` and moved it to `utils.rs`

Issue #381, opened 2024-09-03
Fix EF Tests code reading


belfortep:
There is a bug while reading the codes from the tests, we get the ascii value of the characters instead of getting the hexadecimal value

PR #382, opened 2024-09-03
Ef tests fix

belfortep:
Closes https://github.com/lambdaclass/evm_mlir/issues/381
## Summary
This PR fix an issue with the parsing of the code from the EF
## Changes
- Added a function to parse from the hexadecimal string to the hexadecimal value
- Deleted from ignored groups all workings tests
- Added tests that started to fail
- Can tests code without a `to` address

### DIFF
--- a/tests/ef_tests.rs
+++ b/tests/ef_tests.rs
@@ -56,9 +56,9 @@ fn get_ignored_groups() -> HashSet<String> {
         "stTransitionTest".into(),
         "stCreate2".into(),
         "stSpecialTest".into(),
-        "stSLoadTest".into(),           //
-        "stRecursiveCreate".into(),     //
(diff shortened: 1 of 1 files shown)


lazcanoluca (approved):

lazcanoluca:
> Added tests that started to fail

I don't understand this, are these new tests, or tests that began failing because of the PR's changes?

belfortep:
> > Added tests that started to fail
> 
> I don't understand this, are these new tests, or tests that began failing because of the PR's changes?

Yes, these are tests that began failing because of the changes in the PR

lazcanoluca:
Oh okay I get it! I hadn't noticed the HashSet was returned from the `get_ignored_groups()` function.

JulianVentura (approved):
LGTM

Issue #386, opened 2024-09-04
Fix EF tests transaction data load


belfortep:
It's the same bug as the one while reading the code, we get the ascii value of the characters instead of getting the hexadecimal value

PR #389, opened 2024-09-04
Fix env data tests

belfortep:
Closes https://github.com/lambdaclass/evm_mlir/issues/386
## Summary
This PR fix an issue with the parsing of the transaction data of EF tests
## Changes
- Use of the function to parse from the hex string to hex value
- Updated the list of ignored group of tests
- Small fix to the `call` syscall, using `checked_sub` so it doesn't panic

### DIFF
--- a/tests/ef_tests.rs
+++ b/tests/ef_tests.rs
@@ -3,6 +3,7 @@ use std::{
     path::Path,
 };
 mod ef_tests_executor;
+use bytes::Bytes;
 use ef_tests_executor::models::{AccountInfo, TestSuite};
 use evm_mlir::{db::Db, env::TransactTo, Env, Evm};
 
@@ -31,10 +32,9 @@ fn get_ignored_groups() -> HashSet<String> {
     HashSet::from([
         "stEIP4844-blobtransactions".into(),
         "stEIP5656-MCOPY".into(),
-        "stEIP1153-transientStorage".into(),
         "stEIP3651-warmcoinbase".into(),
-        "stEIP3860-limitmeterinitcode".into(),
         "stArgsZeroOneBalance".into(),
+        "stTimeConsuming".into(),
         "stRevertTest".into(),
         "eip3855_push0".into(),
         "eip4844_blobs".into(),
@@ -52,28 +52,24 @@ fn get_ignored_groups() -> HashSet<String> {
         "stPreCompiledContracts2".into(),
         "stZeroKnowledge2".into(),
         "stDelegatecallTestHomestead".into(),
-        "stTimeConsuming".into(),
         "stEIP150singleCodeGasPrices".into(),
-        "stTransitionTest".into(),
         "stCreate2".into(),
         "stSpecialTest".into(),
+        "stSLoadTest".into(),
+        "stRecursiveCreate".into(),
+        "vmIOandFlowOperations".into(),
         "stEIP150Specific".into(),
-        "eip3651_warm_coinbase".into(),
         "stExtCodeHash".into(),
         "stCallCodes".into(),
         "stRandom2".into(),
         "stMemoryStressTest".into(),
         "stStaticFlagEnabled".into(),
         "vmTests".into(),
-        "opcodes".into(),
         "stEIP158Specific".into(),
         "stZeroKnowledge".into(),
-        "stShift".into(),
         "stLogTests".into(),
-        "eip7516_blobgasfee".into(),
         "stBugs".into(),
         "stEIP1559".into(),
-        "stSelfBalance".into(),
         "stStaticCall".into(),
         "stCallDelegateCodesHomestead".into(),
         "stMemExpandingEIP150Calls".into(),
@@ -82,10 +78,7 @@ fn get_ignored_groups() -> HashSet<String> {
         "stCodeCopyTest".into(),
         "stPreCompiledContracts".into(),
         "stNonZeroCallsTest".into(),
-        "stChainId".into(),
-        "vmLogTest".into(),
         "stMemoryTest".into(),
-        "stWalletTest".into(),
         "stRandom".into(),
         "stInitCodeTest".into(),
         "stBadOpcode".into(),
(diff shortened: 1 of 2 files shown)


tests/ef_tests.rs
```diff
@@ -96,6 +91,10 @@ fn get_ignored_groups() -> HashSet<String> {
         "stExample".into(),
         "vmArithmeticTest".into(),
         "stQuadraticComplexityTest".into(),
+        "stSelfBalance".into(),
+        "stEIP3855-push0".into(),
+        "stWalletTest".into(),
+        "vmLogTest".into(),
```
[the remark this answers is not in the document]
belfortep: Yes, these tests are broken now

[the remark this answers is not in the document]
juanimedone: So they were just randomly succeeding before, right?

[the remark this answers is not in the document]
belfortep: Yes, when they access their data they were getting unexpected things that could or not work

juanimedone (approved):
LGTM!

tests/ef_tests.rs
```diff
@@ -152,7 +151,7 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
             env.tx.caller = sender;
             env.tx.gas_limit = unit.transaction.gas_limit[test.indexes.gas].as_u64();
             env.tx.value = unit.transaction.value[test.indexes.value];
-            env.tx.data = unit.transaction.data[test.indexes.data].clone();
+            env.tx.data = convert_to_hex(unit.transaction.data[test.indexes.data].clone());
```
MegaRedHand, line 154: Can we rename this function? `convert_to_hex` makes me think it turns bytes to their hex representation when it does the opposite.

tests/ef_tests.rs
```diff
@@ -194,15 +193,30 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
+                            return Err("Wrong output".into());
+                        }
+                    }
+                }
+                (
+                    Some(_),
+                    ExecutionResult::Halt {
+                        reason: _,
+                        gas_used: _,
+                    },
```
MegaRedHand, line 211: You can use `..` instead of listing all fields explicitly

tests/ef_tests.rs
```diff
@@ -194,15 +193,30 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
+                        unit.out.as_ref().zip(res.result.output())
+                    {
+                        if expected_output != output {
+                            return Err("Wrong output".into());
+                        }
+                    }
+                }
+                (
+                    Some(_),
+                    ExecutionResult::Halt {
```
MegaRedHand, line 208: Are we sure this doesn't include reverts?

tests/ef_tests.rs
```diff
@@ -194,15 +193,30 @@ fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
-            if test.expect_exception.is_some() {
-                assert!(!res.result.is_success());
-                // NOTE: the expect_exception string is an error description, we don't check the expected error
-                continue;
+            match (&test.expect_exception, &res.result) {
+                (None, _) => {
+                    if let Some((expected_output, output)) =
+                        unit.out.as_ref().zip(res.result.output())
+                    {
+                        if expected_output != output {
```
MegaRedHand, line 201: Isn't this ignoring output mismatches?
I think we can simplify this to:

```suggestion
                    if unit.out.as_ref() != res.result.output() {
```

tests/ef_tests.rs
belfortep: Oh that's true, it takes the hex and converts to decimal. Changed the name and added a little documentation

tests/ef_tests.rs
belfortep: You are right, it should have the `Revert` enum option also

tests/ef_tests.rs
```diff
@@ -34,12 +34,11 @@ fn get_ignored_groups() -> HashSet<String> {
         "stEIP5656-MCOPY".into(),
         "stEIP3651-warmcoinbase".into(),
         "stArgsZeroOneBalance".into(),
-        "stTimeConsuming".into(),
+        //"stTimeConsuming".into(), // this works, but it's REALLY is time consuming
```
juanimedone, line 37:
```suggestion
        //"stTimeConsuming".into(), // this works, but it's REALLY time consuming
```

tests/ef_tests.rs
```diff
@@ -105,8 +105,11 @@ fn get_ignored_suites() -> HashSet<String> {
     ])
 }
 
-fn convert_to_hex(account_info_code: Bytes) -> Bytes {
-    let hex_string = std::str::from_utf8(&account_info_code[2..]).unwrap(); // we don't need the 0x
+/// Receives a Bytes object with the hex representation
+/// And returns a Bytes object with the decimal representation
+/// Taking the hex numbers by pairs
+fn bytes_in_hex_to_decimal(bytes_in_hex: Bytes) -> Bytes {
+    let hex_string = std::str::from_utf8(&bytes_in_hex[2..]).unwrap(); // we don't need the 0x
```
MegaRedHand, line 112: Let's check this is `0x` and fail if not.

tests/ef_tests.rs
```diff
@@ -105,8 +105,11 @@ fn get_ignored_suites() -> HashSet<String> {
     ])
 }
 
-fn convert_to_hex(account_info_code: Bytes) -> Bytes {
-    let hex_string = std::str::from_utf8(&account_info_code[2..]).unwrap(); // we don't need the 0x
+/// Receives a Bytes object with the hex representation
+/// And returns a Bytes object with the decimal representation
+/// Taking the hex numbers by pairs
+fn bytes_in_hex_to_decimal(bytes_in_hex: Bytes) -> Bytes {
```
MegaRedHand, line 111: How about `decode_hex` instead?

tests/ef_tests.rs
belfortep: Should we change to return an Option or just panics?

MegaRedHand (approved):
LGTM

Issue #312, opened 2024-06-28
Cap gas refunds


TomasArrachea:
Add limit to how much gas can be refunded: one fifth of the total transaction cost.
Update all tests that test refunds.

PR #391, opened 2024-09-05
Refund tests

belfortep:
Closes https://github.com/lambdaclass/evm_mlir/issues/312
## Summary
This PR try to fix the stRefundTest and the issue for limiting gas refunds to 1/5 of the total gas used
## Changes
- Added a limit of 1/5 of the total gas used to be refunded
- New tests for this functionality
- Updated old tests for the new gas cap

### DIFF
--- a/src/syscall.rs
+++ b/src/syscall.rs
@@ -40,7 +40,7 @@ use std::collections::HashMap;
 /// Function type for the main entrypoint of the generated code
 pub type MainFunc = extern "C" fn(&mut SyscallContext, initial_gas: u64) -> u8;
 
-pub const GAS_REFUND_FRACTION: u64 = 5;
+pub const GAS_REFUND_DENOMINATOR: u64 = 5;
 
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[repr(C, align(16))]
@@ -196,7 +196,7 @@ impl<'c> SyscallContext<'c> {
         let gas_refunded = self
             .inner_context
             .gas_refund
-            .min(gas_used / GAS_REFUND_FRACTION);
+            .min(gas_used / GAS_REFUND_DENOMINATOR);
         let exit_status = self
             .inner_context
             .exit_status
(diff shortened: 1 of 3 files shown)


src/syscall.rs
```diff
@@ -189,9 +191,12 @@ impl<'c> SyscallContext<'c> {
 
     pub fn get_result(&self) -> Result<ResultAndState, EVMError> {
         let gas_remaining = self.inner_context.gas_remaining.unwrap_or(0);
-        let gas_refunded = self.inner_context.gas_refund;
         let gas_initial = self.env.tx.gas_limit;
         let gas_used = gas_initial.saturating_sub(gas_remaining);
+        let gas_refunded = self
+            .inner_context
+            .gas_refund
+            .min(gas_used / GAS_REFUND_FRACTION);
```
lazcanoluca, line 199: I think it would be a bit clearer to have a dedicated method for computing this, something like `self.get_gas_refund() -> u64`. What do you think?

src/syscall.rs
```diff
@@ -40,6 +40,8 @@ use std::collections::HashMap;
 /// Function type for the main entrypoint of the generated code
 pub type MainFunc = extern "C" fn(&mut SyscallContext, initial_gas: u64) -> u8;
 
+pub const GAS_REFUND_FRACTION: u64 = 5;
```
lazcanoluca, line 43: I think the name is misleading. It says `FRACTION` but the number is actually the denominator. It may be better to rename the constant to make it clearer that it's the denominator. Alternatively, `0.2` could directly be stored as the constant, and then multiplied instead of divided.

belfortep: Oh you are right, it doesn't make any sense saying fraction and being a u64, I like the idea of storing directly the 0.2

belfortep: Changed my mind, naming it denominator seems better

src/syscall.rs
belfortep: Seems a little overkill, as it will only be a function that receives `self` and the `gas_used` just to do the same `.min()`

lazcanoluca: I see, I like methods I guess lol

lazcanoluca (approved):
lgtm 🚀

JulianVentura (approved):
LGTM! 🚀

PR #401, opened 2024-09-06
test: add `ecmul` and `ecpairing` integration tests

juanimedone:
Closes #377 (includes #385 and #387). Comes after #398. Work to be continued in #413.

## Changes
- Fix attempt to copy empty `return_data` to memory.
- Fix `ecadd`'s invalid points unit tests.
- Fix `ecadd`'s integration tests that return `PrecompileError`.
- Add integration tests for `ecmul`.
- Add integration tests for `ecpairing`.

PR #415, opened 2024-09-12
fix: gas consumption on precompile error cases

### LINKED ISSUE #413: Fix gas consumption on precompile failures
Make the gas deduction for those cases in `execute_precompile` instead of doing it inside each precompile function.

Contemplate the empty input cases (valid for some precompiles).

juanimedone:
Closes #413. Comes after #401.

## Summary
Changes how the consumed gas is calculated on precompile failures. Also removes `calldata`'s length check where it doesn't correspond.

## Changes
- Make new `execute_precompile` function handle the gas consumption on `PrecompileError` cases.
- Remove `calldata`'s length check in `ecrecover`, `modexp`, `ecadd` and `ecmul` and use a new auxiliary `right_pad` function to pad it with zeroes when necessary.
- Rewrite empty `calldata` unit tests.
- Add return code and gas consumption checks to all unit tests.


belfortep (approved):
LGTM! 🚀

src/precompiles.rs
```diff
@@ -951,7 +914,6 @@ mod tests {
         let result = ecmul(&calldata, gas_limit, &mut consumed_gas);
 
         assert!(matches!(result, Err(PrecompileError::NotEnoughGas)));
-        assert_eq!(consumed_gas, gas_limit);
```
MegaRedHand, line 954: Why were these checks removed?

juanimedone: I'll make a few adjustments and put them back again.

juanimedone: Done! Now there are both return code and gas consumption checks in all tests.

MegaRedHand (approved):
LGTM

PR #417, opened 2024-09-13
CI parallelization

### LINKED ISSUE #416: Improve CI speed by adding parallelization to EF tests
Currently, the CI pipeline runs sequentially, which results in longer times for the EF tests, especially the stTimeConsuming suite

belfortep:
Closes #416 
## Summary
This PR will change the ci for running the EF tests in 2 different workers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,9 @@ jobs:
   ethereum-tests:
     name: Ethereum Tests (Ubuntu)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        type: [ fast, slow ]
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -143,4 +143,40 @@ jobs:
...
+      - name: Install testing tools
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
+
+      - name: Download test vectors
+        run: make ethtests
+
+      - name: Run ethereum tests
+        run: make test-eth-slow
```
MegaRedHand, line 182: We can use [`strategy.matrix`](https://docs.github.com/es/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) to avoid duplicating this job. Something like:

```yaml
time-consuming-test:
  strategy:
    matrix:
      type: [fast, slow]
  steps:
    - ...
    - if: ${{ matrix.type == 'fast' }}
      run: make test-eth
    - if: ${{ matrix.type == 'slow' }}
      run: make test-eth-slow
```

tests/time_consuming_test.rs
```diff
@@ -0,0 +1,121 @@
+use std::{collections::HashMap, path::Path};
+mod ef_tests_executor;
+use ef_tests_executor::{
+    models::{AccountInfo, TestSuite},
+    test_utils::decode_hex,
+};
+use evm_mlir::{db::Db, env::TransactTo, result::ExecutionResult, Env, Evm};
+
+fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
```
MegaRedHand, line 9: Does this have any change from the one in `ef_tests.rs`? If not, we can move `ef_tests::run_test` to a module and call it from both files.

MegaRedHand (changes requested):

tests/time_consuming_test.rs
belfortep: Yes, this `run_test` have some little differences, the main one is that the `ef_tests.rs` one ignore groups of tests (including the time_consuming) and this not, but we can move most of the logic to a module as it's the same

.github/workflows/ci.yml
belfortep: Didn't know about that, taking a look 👀

tests/time_consuming_test.rs
MegaRedHand: Let's extract the test running logic to a module, then.

MegaRedHand (approved):
LGTM

juanimedone (approved):

PR #419, opened 2024-09-13
Solidity tests

### LINKED ISSUE #107: General test (E2E) for all the implemented opcodes
We need to have a general test that tests the correct functionality of all the implemented opcodes at the same time. 
This should be done in solidity.

belfortep:
Closes #107 
## Summary
Adds solidity contracts, their compiled bytecode and tests for this contracts
The contracts are `Factorial`, `Fibonacci` and a recursive version of Fibonacci, `RecursiveFibonacci`

## Changes
- Updated the CI to download the Solc compiler and to compile the contracts
- Added a new test file that loads the contracts `.bin` and executes them


Makefile
```diff
@@ -106,3 +106,9 @@ ethtests: ${ETHTEST_TAR}
 	tar -xzmf "$<" --strip-components=1 -C "$@"
 	@cat ${ETHTEST_SHASUM}
 	sha256sum -c ${ETHTEST_SHASUM}
+
+install-solc:
+	brew install solidity
+
+compile-solidity:
```
lazcanoluca, line 113: Maybe call this:

```suggestion
compile-solidity-examples:
```

Just to make it very explicit that it compiles the _examples_.

lazcanoluca: Well, the _tests_ I should say...

programs/Fibonacci.bin
```diff
@@ -0,0 +1 @@
...801561000f575f5ffd5b5060043610610034575f3560e01c806361047ff4146100385780636537214714610068575b5f5ffd5b610052600480360381019061004d919061012e565b610086565b60405161005f9190610168565b60405180910390f35b6100706100f2565b60405161007d9190610168565b60405180910390f35b5f5f8203610096575f90506100ed565b600182036100a757600190506100ed565b5f5f90505f600190505f5f600290505b8581116100e55782846100ca91906101ae565b915082935081925080806100dd906101e1565b9150506100b7565b508093505050505b919050565b5f5481565b5f5ffd5b5f819050919050565b61010d816100fb565b8114610117575f5ffd5b50565b5f8135905061012881610104565b92915050565b5f60208284031215610143576101426100f7565b5b5f6101508482850161011a565b91505092915050565b610162816100fb565b82525050565b5f60208201905061017b5f830184610159565b92915050565b7f4e487b71000000000000000000000000000000000000000000000000000000005f52601160045260245ffd5b5f6101b8826100fb565b91506101c3836100fb565b92508282019050808211156101db576101da610181565b5b92915050565b5f6101eb826100fb565b91507fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff820361021d5761021c610181565b5b60018201905091905056fea2646970667358221220cc7e80883e46c9b1dcc50e00d58d2ebc200d477cea82d9c2fc35d1ee0999b9ce64736f6c634300081b0033
```
lazcanoluca, line 1: Is it necessary to have the binaries? Can't they be built when, or before, running the solidity tests? I imagine the scenario where we modify the solidity contract, but forget to update the binaries. Instead, if they are built when needed, there is only one source of truth for the tests. What do you think?

lazcanoluca (approved):
I left a couple of thoughts you may consider, approving either way 🚀

Makefile
belfortep: the best of both worlds, test-examples 😎

lazcanoluca: `compile-solidity-test-examples-that-test-compilation-and-are-actually-example-contracts`

belfortep: 🤣🤣🤣

programs/Fibonacci.bin
belfortep: You are right, the binaries shouldn't be added directly in the repo, tried to be lazy not changing the CI but we should do it that way, changing that

lazcanoluca (approved):

README.md
```diff
@@ -128,6 +128,26 @@ make ethtests   # downloads the tests
 make test-eth
 ```
 
+To run the solidity tests if you don't have the compiled binaries, you can run
+
+```bash
+make compile-solidity
```
iinaki, line 134:
```suggestion
make compile-solidity-test-examples
```

iinaki, line 134:
```suggestion
make compile-solidity-test-examples
```

igaray (approved):

PR #434, opened 2024-09-18
Add docs to precompiles constants

iinaki:


src/constants.rs
juanimedone, line 1: Probably it should be better to place the description of each precompile right before each function in `precompiles.rs`. You can use `///`.

juanimedone: Yes, good idea.

iinaki: Done!

src/precompiles.rs
```diff
@@ -159,6 +172,8 @@ pub fn modexp(
     Ok(Bytes::copy_from_slice(output))
 }
 
+/// Point addition on the elliptic curve 'alt_bn128'.
```
juanimedone, line 175:
```suggestion
/// Point addition on the elliptic curve 'alt_bn128' (also referred as 'bn254').
```

src/precompiles.rs
```diff
@@ -217,6 +232,8 @@ pub fn ecadd(
     Ok(Bytes::from(res))
 }
 
+/// Scalar multiplication on the elliptic curve 'alt_bn128'.
```
juanimedone, line 235:
```suggestion
/// Scalar multiplication on the elliptic curve 'alt_bn128' (also referred as 'bn254').
```

src/precompiles.rs
```diff
@@ -257,6 +274,11 @@ pub fn ecmul(
     Err(PrecompileError::InvalidEcPoint)
 }
 
+/// Elliptic curve pairing operation required in order to perform zkSNARK verification within the block gas limit. Bilinear function on groups on the elliptic curve “alt_bn128”.
```
juanimedone, line 277:
```suggestion
/// Elliptic curve pairing operation required in order to perform zkSNARK verification within the block gas limit. Bilinear function on groups on the elliptic curve 'alt_bn128' (also referred as 'bn254').
```

juanimedone (approved):
Just added a little detail, the rest LGTM.

src/precompiles.rs
iinaki: Done!

src/precompiles.rs
iinaki: Done!

juanimedone:
Now the PR's title can just be 'Add docs to precompiles' or 'Add precompiles' docs'.

igaray (approved):

PR #436, opened 2024-09-18
Refactor constant operations into const functions

iinaki:
- Noticed that in the `ecpairing` precompile some operations could be moved to constant functions in order to optimize a bit the code


igaray (approved):

src/precompiles.rs
```diff
@@ -28,6 +28,8 @@ use num_bigint::BigUint;
 use secp256k1::{ecdsa, Message, Secp256k1};
 use sha3::{Digest, Keccak256};
 
+/// ECDSA public key recovery function.
+/// More info in https://eips.ethereum.org/EIPS/eip-2, https://eips.ethereum.org/EIPS/eip-1271 and https://www.evm.codes/precompiled.
```
lazcanoluca, line 32: It looks like the comment got repeated here

lazcanoluca (approved):

src/precompiles.rs
iinaki: done!

PR #437, opened 2024-09-18
Transaction validation

### LINKED ISSUE #435: Transaction validation
Transactions are lacking validation before execution. Many EF tests fail because of this. Transaction validation should be implemented, with testing.

lazcanoluca:
Working branch for transaction validation.

Closes #435.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -44,6 +44,10 @@ impl Db {
         Self::default()
     }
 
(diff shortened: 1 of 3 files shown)


lazcanoluca:
At this point, all of the validations are implemented. It may be necessary to adapt the unit tests so they don't fail on validation.

PR #438, opened 2024-09-19
Fix eip3860-initcode from EF

### LINKED ISSUE #376: Fix stEIP3860 from EF

belfortep:
Closes #376 

## Summary
Tests that returns error when running the `evm.transact()` are now handled, fixed an error with conversion of types resulting on wrong gas consumption with Create transactions

## Changes
- Updated the tests verification to catch error when the initcode throws an exception but is a expected one
- Changed the calculation of `init_code_cost`


tests/ef_tests_executor/test_utils.rs
```diff
@@ -156,10 +165,17 @@ pub fn run_test(path: &Path, contents: String) -> datatest_stable::Result<()> {
...
-            let res = evm.transact().unwrap();
-            verify_result(test, unit.out.as_ref(), &res.result)?;
+            let res = evm.transact();
+            verify_result(test, unit.out.as_ref(), &res)?;
             // TODO: use rlp and hash to check logs
-            verify_storage(&test.post_state, res.state);
+            match res {
+                Ok(res) => {
+                    verify_storage(&test.post_state, res.state);
+                }
```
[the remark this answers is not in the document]
belfortep: Oops, I always forget about `if let` and use `match`

tests/ef_tests_executor/test_utils.rs
```diff
@@ -97,23 +97,32 @@ fn setup_evm(test: &Test, unit: &TestUnit) -> Evm<Db> {
...
+            Ok(()) //Got error and expeted one
         }
-        _ => Err("Expected exception but got none".into()),
+        (Some(_), Ok(execution_result)) => {
+            match execution_result.result {
+                ExecutionResult::Halt { .. } | ExecutionResult::Revert { .. } => {
+                    Ok(()) // Got error and got expected halt/revert
+                }
+                _ => Err("Expected exception but got none".into()),
+            }
```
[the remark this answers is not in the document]
belfortep: Where should we use `map_err` here?

src/constants.rs
```diff
@@ -121,8 +121,9 @@ pub mod gas_cost {
     pub const TX_ACCESS_LIST_STORAGE_KEY_COST: u64 = 1900;
     pub const MAX_CODE_SIZE: usize = 0x6000;
 
-    pub fn init_code_cost(init_code_length: usize) -> u64 {
-        INIT_WORD_COST as u64 * (init_code_length as u64 + 31) / 32
+    pub fn init_code_cost(init_code_length: u64) -> u64 {
```
lazcanoluca, line 124: I don't quite understand where does the 31 comes from. I looked a bit at the EIP but I don't get it.

lazcanoluca, line 124: The EIP mentions:
> If length of transaction data (initcode) in a create transaction exceeds MAX_INITCODE_SIZE, transaction is invalid.

Should this check be done here, or is it assumed that `init_code_length` is already checked?

If the assumption is that `init_code_length` is < `MAX_INITCODE_SIZE`, an `assert` wouldn't hurt, just to enforce it.

Otherwise (if `init_code_length` can be > `MAX_INITCODE_SIZE`), I think the function should gracefully handle this case with `Result`.

lazcanoluca, line 124: Finally, I think we should add a doc comment (with a link to the EIP), and a couple of unit tests.

lazcanoluca (changes requested):

src/constants.rs
belfortep: to be honest, I don't really know where it comes from, I just accepted the mistery.
Probably is something that acts as a padding for initcode data that isn't a full 32-bytes word

belfortep: Yes, the `init_code_length` is verified in the `validate_transaction` function, we could add the `assert` for an extra verification for a calm of mind

belfortep: You are right, adding some test for the correct gas usage of Create transactions and some docs

lazcanoluca: cool!

lazcanoluca: awesome 🚀

TomasArrachea: This operation computes the amount of 32-length chunks. The operation `int((x + 31) / 32) ` is analogous to `ceil(x / 32)`

tests/ef_tests_executor/test_utils.rs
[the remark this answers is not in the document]
MegaRedHand: I must've mistaken this to be a `match` on `Ok`/`Err` 🤔 
Let's leave it as-is.

MegaRedHand (approved):
LGTM

src/constants.rs
lazcanoluca: I guess [.div_ceil()](https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil) could also be used.

lazcanoluca (approved):

PR #441, opened 2024-09-19
Fix stprecompiledcontracts

juanimedone:


iinaki (approved):

.github/workflows/ci.yml
```diff
@@ -152,3 +152,7 @@ jobs:
       - name: Run slow ethereum tests
         if: ${{ matrix.type == 'slow' }}
         run: make test-eth-slow
+        
+      - name: Run blake2B ethereum test
+        if: ${{ matrix.type == 'blake2B' }}
+        run: make test-eth-blake2B
```
TomasArrachea, line 158: why are this tests running separately?

juanimedone: Because that specific test (`blake2B`) takes like 19 minutes to run, that's why I created a specific Actions step just for it.

TomasArrachea: ok perfect, maybe we leave a comment stating this just for clarification

juanimedone: I left this comment in `tests/ef_tests.rs`:
```rust
fn get_ignored_suites() -> HashSet<String> {
    HashSet::from([
        "ValueOverflow".into(),      // TODO: parse bigint tx value
        "ValueOverflowParis".into(), // TODO: parse bigint tx value
        "blake2B".into(), // this is tested in the blake2B binary because takes a lot of time
    ])
}
```
Is that ok or should I add another one somewhere else?

